PR #1, opened 2017-01-05
Standalone repo config, yarn, circleci integration, shipit integration

GoodForOneFare:
See title.


lemonmade:
Same comments as other repos, then :+1:

PR #4, opened 2017-01-16
Update eslint plugins

ismail-syed:
* Added [`eslint-index`](https://github.com/wagerfield/eslint-index)  package to help improve scripts to help find omitted lint rules
* Added `yarn run rules-status` and `yarn run rules-omitted` scripts (output shown below) 
* Removed `eslint-find-rules` package and script
* Updated elint-plugins `eslint-plugin-flowtype`, `eslint-plugin-mocha`, `eslint-plugin-promise`, `eslint-plugin-react` packages
* Added new rules for `eslint-plugin-react` and `eslint-plugin-promise`

screen shot 2017-01-16 at 1 46 35 pm
screen shot 2017-01-16 at 1 46 50 pm

**Why `lodash/no-single-chain` and `lodash/prefer-chain` are considered 'omitted'.**
>[Joined prefer-chain and no-single-chain into single rule, chaining that also enables disallowing chaining. (685e9e3)](https://github.com/wix/eslint-plugin-lodash/blob/master/CHANGELOG.md#200---2016-10-13)

**Why babel stuff comes up:** 
Babel has no CHANGELOG.md, which could be why this isn't getting picked up properly?  However, they list it as deprecated on their [README.md](https://github.com/babel/eslint-plugin-babel)


lib/config/rules/promise.js
```diff
@@ -1,13 +1,33 @@
...
+  'promise/no-promise-in-callback': 'warn',
+  // Avoid calling cb() inside of a then() (use nodeify] instead)
+  'promise/no-callback-in-promise': 'warn',
+   // Avoid creating new promises outside of utility libs (use pify instead)
+  'promise/avoid-new': 'warn',
+
+
+  // Async/Await Rules
+
+  // Prefer await to then() for reading Promise values
```
minasmart, line 29: Are we using the babel compiler for this or something else? Their implementation is not what I'd call light weight.

ex:
```javascript
async function doTheThing() {
  const whatever = await fetch('google.com');
  
  return whatever;
}
```

transforms to:

```javascript

var doTheThing = function () {
  var _ref = _asyncToGenerator(regeneratorRuntime.mark(function _callee() {
    var whatever;
    return regeneratorRuntime.wrap(function _callee$(_context) {
      while (1) {
        switch (_context.prev = _context.next) {
          case 0:
            _context.next = 2;
            return fetch('google.com');

          case 2:
            whatever = _context.sent;
            return _context.abrupt('return', whatever);

          case 4:
          case 'end':
            return _context.stop();
        }
      }
    }, _callee, this);
  }));

  return function doTheThing() {
    return _ref.apply(this, arguments);
  };
}();

function _asyncToGenerator(fn) { return function () { var gen = fn.apply(this, arguments); return new Promise(function (resolve, reject) { function step(key, arg) { try { var info = gen[key](arg); var value = info.value; } catch (error) { reject(error); return; } if (info.done) { resolve(value); } else { return Promise.resolve(value).then(function (value) { step("next", value); }, function (err) { step("throw", err); }); } } return step("next"); }); }; }
```
[source](http://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=es2015%2Ces2015-loose%2Ces2016%2Ces2017%2Clatest%2Creact%2Cstage-0%2Cstage-1%2Cstage-2%2Cstage-3&experimental=false&loose=false&spec=false&code=async%20function%20doTheThing()%20%7B%0A%20%20const%20whatever%20%3D%20await%20fetch('google.com')%3B%0A%20%20%0A%20%20return%20whatever%3B%0A%7D&playground=true)

It also creates a blocking loop with the `while(1)` which is kinda scary. It does look like there's a defined exit condition, but it's hard to say from the generated code how many iterations it'll block on.

ismail-syed: :flushed: That does look quite scary. 
@GoodForOneFare: Should we keep this as a 'warn', or even have it turned 'off'?

minasmart: Given the blocking behaviour, is it even advantageous to encourage the use of await? I feel like it produce a specific behaviour that may be desirable only in specific circumstances (where blocking IO is required)

tessalt: yeah i have strong feels about preferring await with its current browser support. I'd like to hear some of the justification behind this.

GoodForOneFare: @ismail-syed is just getting started on FED, so these are just his best guesses.  Agree that this should be `off` until browser support is better.

lemonmade: Yup, agreed to turn these two off 👍

mikkoh: Agreed. I think it's too early for `async` `await` at this point.

lib/config/rules/react.js
```diff
@@ -47,6 +49,8 @@ module.exports = {
   'react/prop-types': 'warn',
   // Prevent missing React when using JSX
   'react/react-in-jsx-scope': 'error',
+  // Enforce a defaultProps definition for every prop that is not a required prop
+  'react/require-default-props': 'error',
```
lemonmade: IMO this rule is not worth it, I think unset values are fine for boolean props (which would default to falsey values anyways), and I kind of prefer specifying defaults in actual JS rather than `defaultProps`.

lib/config/rules/promise.js
```diff
@@ -1,13 +1,33 @@
...
   // Enforce standard parameter names for Promise constructors.
   'promise/param-names': 'warn',
   // Ensure that Promise is included fresh in each file instead of relying on the existence of a native promise implementation.
   'promise/no-native': 'off',
+  // Avoid nested .then() or .catch() statements
+  'promise/no-nesting': 'warn',
+  // Avoid using promises inside of callbacks
+  'promise/no-promise-in-callback': 'warn',
+  // Avoid calling cb() inside of a then() (use nodeify] instead)
+  'promise/no-callback-in-promise': 'warn',
```
lemonmade: I feel that this rule and the one above are overly strict, I think there are too many valid cases to keep these on.

lib/config/rules/promise.js
```diff
@@ -1,13 +1,33 @@
...
+   // Avoid creating new promises outside of utility libs (use pify instead)
+  'promise/avoid-new': 'warn',
```
lemonmade: Definitely turn this one off, we definitely want people to be free to use `new Promise` whenever they like.

lemonmade (changes requested):

lib/config/rules/promise.js
ismail-syed: Thanks everyone. 
The two await rules are `off` now. 
70af7aa

lib/config/rules/promise.js
ismail-syed: 👍  
70af7aa.

GoodForOneFare:
GTG once Chris' comments are addressed.  Thank you :)

GoodForOneFare:
Oh, could you start a `CHANGELOG.md`, please?  The [old JS repo's log has plenty of examples for eslint config bumps](https://github.com/Shopify/javascript/blob/v15.0.3/CHANGELOG.md).

wagerfield:
@ismail-syed the omitted rules you are seeing for babel are there because the plugin has not properly declared it's deprecated rules as being deprecated using ESLint's meta object. Even though they have documented these rules as being deprecated in their README, there is no programatic way of `eslint-index` deriving that these rules are deprecated.

[See here for more information](http://eslint.org/docs/developer-guide/working-with-rules#rule-basics)

To remedy this, I have opened an [issues with eslint-plugin-babel](https://github.com/babel/eslint-plugin-babel/issues/118), so hopefully when these fixes are made—you should no longer see these omitted rules for the babel plugin.

As for the 2 lodash rules, that's got me a little stumped. Are you using the latest version of this plugin? I had a [search in the repo](https://github.com/wix/eslint-plugin-lodash/search?utf8=%E2%9C%93&q=prefer-chain) and from the change log it looks like these 2 rules were merged into a new `chaining` rule. The search didn't return any other matches for these 2 rules, so I'm struggling to understand why they are appearing in your omitted list. 

Could you please confirm that you have the latest version of `eslint-plugin-lodash`?

lib/config/rules/promise.js
```diff
@@ -1,13 +1,33 @@
...
   // Ensure that inside a then() you make sure to return a new promise or value.
   'promise/always-return': 'error',
   // Enforce standard parameter names for Promise constructors.
   'promise/param-names': 'warn',
   // Ensure that Promise is included fresh in each file instead of relying on the existence of a native promise implementation.
   'promise/no-native': 'off',
+  // Avoid nested .then() or .catch() statements
+  'promise/no-nesting': 'warn',
+  // Avoid using promises inside of callbacks
+  'promise/no-promise-in-callback': 'warn',
```
minasmart: When working with any of the node API async methods, it's kind of hard to not do this, if you are using any other library that uses promises.

ismail-syed:
@wagerfield That's what I presumed was the reason behind those Babel rules showing up as omitted. Thanks for the clarification. 

Regarding `eslint-plugin-lodash`, I just updated and re-tested locally. An update was sent out today, so a new rule was introduced: `import-scope`. 
I'm on version `eslint-plugin-lodash` version `2.3.1`.

The two rules `lodash/no-single-chain` and `lodash/prefer-chain` still show up.

wagerfield:
@ismail-syed I will do some further investigations as to why these lodash rules are being outputted.

lib/config/rules/promise.js
hzoo: Yeah, not using async also would help with removing `regeneratorRuntime`. Are y'all using babel-polyfill as well?

Alternatives: you can try https://github.com/babel/kneden although needs work since it's not stable yet, https://github.com/MatAtBread/fast-async, or use http://babeljs.io/docs/plugins/transform-async-to-generator if you support browsers that have generators natively. Would recommend https://github.com/babel/babel-preset-env/ for that as well.

Sorry for the random comment haha

lib/config/rules/promise.js
ismail-syed: Fixed with 3b6f088

lib/config/rules/react.js
ismail-syed: Fixed with 3b6f088.

lib/config/rules/promise.js
lemonmade: @hzoo All of our projects support at least IE11, so we're still stuck without native generators, and I'd rather way for async-await until we can do so based on native generators, rather than any kind of runtime. We're excited to start using `babel-preset-env` (https://github.com/Shopify/babel-preset-shopify/issues/4), though. Thanks very much for all your hard work on the Babel project!

ismail-syed:
Thanks @wagerfield, appreciate it! 💯

lemonmade (approved):

GoodForOneFare (approved):

wagerfield:
@ismail-syed FYI https://github.com/babel/eslint-plugin-babel/pull/119

ismail-syed:
Thanks @wagerfield 👍

PR #16, opened 2017-02-21
jquery-dollar-sign-reference now checks LogicalExpression values

### LINKED ISSUE #2: jquery-dollar-sign-reference shold flag jQuery assignments inside LogicalExpressions
The following code should be flagged by the linter as requiring `var $foo`:

```js
var foo = bar || $('baz');
```

/cc @lemonmade

GoodForOneFare:
## What does it do?

Flags `$`-prefix assignment mismatches within LogicalExpressions.  e.g.,

The following code is flagged because the receiving var `$foo` could accept a value from the non-$-prefixed `qux` var.

```js
const $foo = $bar || qux;
```

The following code is flagged because the receiving var `foo` could accept a value from the $-prefixed `$bar` var.

```js
const foo = $bar || qux;
```

-----

## How does it do it?
Updates `checkForValidjQueryReference` to dig into the `right` side of an assignment to find all possible assignment sources.  The list of source nodes is checked for `$`/regular types, and a reworded version of the original logic is applied to the results.

Fixes #2

### DIFF
--- a/lib/rules/jquery-dollar-sign-reference.js
+++ b/lib/rules/jquery-dollar-sign-reference.js
@@ -126,15 +126,34 @@ module.exports = {
       };
     }
 
+    function getValueNodes(node) {
+      var results = [];
+      if (node.type === 'LogicalExpression') {
+        results.push.apply(results, getValueNodes(node.left));
+        results.push.apply(results, getValueNodes(node.right));
+      } else if (node.type === 'ConditionalExpression') {
+        results.push.apply(results, getValueNodes(node.consequent));
+        results.push.apply(results, getValueNodes(node.alternate));
+      } else {
+        results.push(node);
+      }
+      return results;
+    }
+
     function checkForValidjQueryReference(node, left, right) {
       if (right == null) { return; }
       var isjQueryRef = isjQueryReference(left);
-      var isjQueryVal = isjQueryValue(right);
-      // console.log(right);
+      var valueNodeTypes = getValueNodes(right).map(isjQueryValue);
+      var hasDefinitejQueryValue = valueNodeTypes.some(function(nodeType) {
+        return nodeType.definite;
+      });
+      var hasRegularValue = valueNodeTypes.some(function(nodeType) {
+        return !nodeType.possible;
+      });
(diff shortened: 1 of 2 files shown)


lib/rules/jquery-dollar-sign-reference.js
```diff
@@ -126,15 +126,34 @@ module.exports = {
...
+      return results;
+    }
+
     function checkForValidjQueryReference(node, left, right) {
       if (right == null) { return; }
       var isjQueryRef = isjQueryReference(left);
-      var isjQueryVal = isjQueryValue(right);
-      // console.log(right);
+      var valueNodeTypes = getValueNodes(right).map(isjQueryValue);
+      var hasDefinitejQueryValue = valueNodeTypes.some(function(nodeType) {
```
ismail-syed, line 147: This could be a preference thing, but what's our stance on utilizing arrow functions?
```
var hasDefinitejQueryValue = valueNodeTypes.some(nodeType => nodeType.definite);
var hasRegularValue = valueNodeTypes.some(nodeType => !nodeType.possible);
```

ismail-syed (approved):
LGTM 👍

lib/rules/jquery-dollar-sign-reference.js
GoodForOneFare: I'd love to, but the [eslint plugin targets es5](https://github.com/Shopify/eslint-plugin-shopify/blob/master/package.json#L35), and doesn't have a transpilation step.

I'm guessing this is because:
- We don't want to restrict our plugin to people using the latest & greatest node version
- We don't want to introduce a build step for a small project / just never had the time

ismail-syed: Ahhh, okay.

nathanmarks:
@GoodForOneFare FYI, when I install deps my `yarn.lock` gets updated.

nathanmarks (approved):
Checked it out and threw a few more things at it for fun. Didn't break.

GoodForOneFare:
Thanks, folks!

PR #19, opened 2017-03-06
Eslint upgrade

GoodForOneFare:
Bumps the `eslint` version.  This is to get `shopify/shopify` / `sewing-kit` / `eslint-plugin-shopify` all on compatible eslint dependencies.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,3 +24,7 @@
 
 ### Removed
 - Removed `eslint-find-rules` package ([#4](https://github.com/Shopify/eslint-plugin-shopify/pull/4))
+
+# Pre-15.1.1 Changelog
+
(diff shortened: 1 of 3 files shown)


ismail-syed:
Are we not gonna add all the new rules that were released between the `3.10.x` -> `3.17.x`.
Or is that PR for another day?

ismail-syed:
Shouldn't we update the peer dependency as well [51](https://github.com/Shopify/eslint-plugin-shopify/blob/446ec9a09f5d75b0023e0842319cf390cc7a8695/package.json#L51).

GoodForOneFare:
@ismail-syed updated the peer dependency.  Nice one 👍 

This PR is to unblock sewing-kit/reports, so adding new rules will have to happen another day.

ismail-syed (commented):
:shipit: 
👍

PR #23, opened 2017-03-28
Specify the minor version tested against

leo-shopify:
Clients using earlier versions could potentially experience breakage.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "eslint-plugin-shopify",
-  "version": "15.3.0",
+  "version": "15.3.1",
   "description": "Shopify’s ESLint rules and configs.",
   "keywords": [
     "eslint",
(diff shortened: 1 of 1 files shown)


ismail-syed:
@leo-shopify what's the breakage? Just curious.

leo-shopify:
It's not that something is broken right now but that my previous merge 1ba1dd064cc0fc0f702e5360f41cb356da2aa895 by being too lenient with the minor version could potentially cause trouble.
e.g.: if a downstream user uses a version of eslint prior to 3.17, the `3.x` range would have accepted it as valid even if *eslint-plugin-shopify* requires features that the user doesn't have.
i.e.: minor versions promise *backwards* compatibility, but not *forward*.
Debugging this kind of issues would be a nightmare and I wanted to avoid it altogether.

ismail-syed (approved):

leo-shopify:
Thanks!
Please lmk if I need to do anything to publish the new version

PR #30, opened 2017-05-29
Turn off prefer-destructuring

ismail-syed:


lemonmade (approved):

GoodForOneFare:
Was this causing problems?

lemonmade:
Yes, it seemed to want you to destructure at every level of dot notation in an object, and was doing the wrong thing for numbered keys (wanting you to object-destructure them), and seemingly wanted it for places it is invalid, like using keys that are just variables or for keys that are strings that would not make valid identifiers.

PR #36, opened 2017-08-04
Allow `$foo = await bar()`

GoodForOneFare:
See title.

### DIFF
--- a/tests/lib/rules/jquery-dollar-sign-reference.js
+++ b/tests/lib/rules/jquery-dollar-sign-reference.js
@@ -234,6 +234,15 @@ ruleTester.run('jquery-dollar-sign-reference', rule, {
 
     {code: 'var foo = {["bar"]: "bar"}', parserOptions: {ecmaVersion: 2015}},
     {code: 'var foo = {["$bar"]: $bar}', parserOptions: {ecmaVersion: 2015}},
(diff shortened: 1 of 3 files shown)


tests/lib/rules/jquery-dollar-sign-reference.js
```diff
@@ -231,6 +231,11 @@ ruleTester.run('jquery-dollar-sign-reference', rule, {
     {code: 'var foo = bar || qux;'},
     {code: 'var foo = bar && quz;'},
     {code: 'var foo = bar || (qux && baz);'},
+
+    {code: 'async function foo() { var foo = await aPromise; }', parserOptions: {ecmaVersion: 2017}},
+    {code: 'async function foo() { var foo = await $aPromise; }', parserOptions: {ecmaVersion: 2017}},
+    {code: 'async function foo() { var $foo = await aPromise; }', parserOptions: {ecmaVersion: 2017}},
+    {code: 'async function foo() { var $foo = await $aPromise; }', parserOptions: {ecmaVersion: 2017}},
```
lemonmade: Might want to test this for a few assignment expressions and some awaited things that are function calls, just to cover all the other cases.

lemonmade (approved):

PR #43, opened 2017-09-21
Utilize asyncArrow option for space-before-function-paren

ismail-syed:


GoodForOneFare (approved):
Merci! 💃

GoodForOneFare:
Reason for this change: we sifted through `web`, `sewing-kit`, and `shopify/shopify` code and async functions are _always_ prefixed with a space in those codebases.

This change just formalizes what's already there.

PR #44, opened 2017-09-27
lines-around-directive deprecated since ESLint 4.0.0

ismail-syed:
[lines-around-directive](https://eslint.org/docs/rules/lines-around-directive) has been deprecated as per their rule docs. I didn't/don't see it explicitly listed in their changelog. It's been replaced with replaced by the [`padding-line-between-statements`](https://eslint.org/docs/rules/padding-line-between-statements) rule, which we have [disabled](https://github.com/Shopify/eslint-plugin-shopify/blob/27de727770f287c22f053bda7364d91e225a2157/lib/config/rules/stylistic-issues.js#L139).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Changed
 * `space-before-function-paren` now uses `asyncArrow` option (eg. `async () => {}`)
-
+* `lines-around-directive` was deprecated in ESLint `v4.0.0`.
 ## [17.1.0] - 2017-09-19
 
 ### Added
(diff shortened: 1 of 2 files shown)


GoodForOneFare:
This was replaced by [`padding-line-between-statements`](https://eslint.org/docs/rules/padding-line-between-statements).  Can you try adding a `directive` config that mimics the deprecated rule.

The docs suggest this:
```
[{ blankLine: "always", prev: "directive", next: "*" }, { blankLine: "any", prev: "directive", next: "directive" }]
```

ismail-syed:
@GoodForOneFare another 👀 , please.

GoodForOneFare (approved):
👍 Tried this out in admin (with an eslint version bump) and it works as expected.

PR #45, opened 2017-10-03
🔥 class-methods-use-this

davidcornu:
https://github.com/Shopify/javascript/issues/263

### DIFF
--- a/lib/config/rules/best-practices.js
+++ b/lib/config/rules/best-practices.js
@@ -8,7 +8,7 @@ module.exports = {
   // Treat var statements as if they were block scoped
   'block-scoped-var': 'error',
   // Enforce that class methods utilize this
-  'class-methods-use-this': 'error',
+  'class-methods-use-this': 'off',
   // Specify the maximum cyclomatic complexity allowed in a program
   complexity: 'off',
   // Require return statements to either always or never specify values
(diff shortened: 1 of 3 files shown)


GoodForOneFare:
FYI, Chris is on vacation until the 17th.

davidcornu:
@GoodForOneFare am not in any particular hurry 😛

lib/config/rules/best-practices.js
```diff
@@ -7,8 +7,6 @@ module.exports = {
   'array-callback-return': 'error',
   // Treat var statements as if they were block scoped
   'block-scoped-var': 'warn',
-  // Enforce that class methods utilize this
-  'class-methods-use-this': 'warn',
```
lemonmade: Can we please keep the rule but just set it to `off`?

lemonmade (approved):
Just switch to explicitly off versus excluding the rule, then we're good to go 👍

lib/config/rules/best-practices.js
davidcornu: Sure thing

GoodForOneFare:
Please update the CHANGELOG, too.

davidcornu:
Changed to `off` and updated CHANGELOG.

PR #46, opened 2017-10-12
Add a prettier config

ismail-syed:
* Created and exposed a prettier config that can be consumed for JS projects via `plugin:shopify/prettier`.
* Formatted this repo with the new config

This PR also adds the following dependencies: 
* [prettier](https://github.com/prettier/prettier): Core prettier formatter. 
* [eslint-plugin-prettier](https://github.com/prettier/eslint-plugin-prettier): Exposes prettier as an eslint rule
* [eslint-config-prettier](https://github.com/prettier/eslint-config-prettier): A maintained config that has conflicting rules [disabled](https://github.com/prettier/eslint-config-prettier/blob/master/index.js). 

Ran this against core, [here's](https://github.com/Shopify/shopify/commit/3700835f55926b6064904e11571d8ab70e6929f3) a diff to get an idea of potential changes. Let me know if you see any concerns.


lib/rules/restrict-full-import.js
```diff
@@ -1,66 +1,76 @@
...
-  create: function(context) {
-    var restricted = context.options[0] || [];
+  create(context) {
+    const restricted = context.options[0] || [];
 
     function isRestrictedModule(mod) {
       return restricted.indexOf(mod) >= 0;
     }
 
+    /* eslint-disable prettier/prettier*/
```
ismail-syed: `eslint --fix` with the prettier/prettier config breaks this. It outputs invalid code. Create a reproducible issue in issue in `eslint-plugin-prettier`.

lib/rules/sinon-prefer-meaningful-assertions.js
```diff
@@ -47,53 +47,72 @@ var SINON_CHAI = [
...
+      }
+      return (
+        (node.type === 'MemberExpression' &&
+          methods.indexOf(node.property.name) >= 0) ||
+        (node.type === 'CallExpression' &&
+          isRewritableSinonExpression(node.callee, methods))
+      );
     }
 
+    /* eslint-disable prettier/prettier*/
```
ismail-syed: Same here, `eslint --fix` with the prettier/prettier config breaks this. It outputs invalid code. Create a reproducible issue in issue in `eslint-plugin-prettier`.

tests/.eslintrc.js
```diff
@@ -2,5 +2,6 @@ module.exports = {
   extends: 'plugin:shopify/esnext',
   rules: {
     'import/no-extraneous-dependencies': 'off',
+    'prettier/prettier': 'off',
```
ismail-syed: Alternatively, we could extend the `plugin:shopify/prettier` config and run prettier against the `/tests` code.

package.json
```diff
@@ -18,6 +18,8 @@
     "rules-status": "eslint-index lib/config/all.js  --format table",
     "rules-omitted": "eslint-index lib/config/all.js  --status omitted",
     "lint": "eslint . --max-warnings 0",
+    "format": "prettier --write --single-quote --no-bracket-spacing --trailing-comma es5 './**/*.js'",
+    "eslint-check": "eslint --print-config .eslintrc.js | eslint-config-prettier-check",
```
ismail-syed: As we update `prettier` and/or `eslint` hopefully, this can continue to list conflicting rules.

lib/rules/restrict-full-import.js
ismail-syed: I can confirm that this is a bug. 
See https://github.com/prettier/eslint-config-prettier/issues/31 & https://github.com/prettier/eslint-plugin-prettier/issues/65

lib/rules/sinon-prefer-meaningful-assertions.js
ismail-syed: See prettier/eslint-config-prettier#31

lib/rules/restrict-full-import.js
lydell: Instead of disabling prettier here, you can run `eslint --fix`, manually correct the syntax error, and the `eslint --fix` again.

ismail-syed: Thanks for the tip. 
I'd prefer leaving as is. I've commented a link to the issue for context.

package.json
```diff
@@ -18,6 +18,7 @@
     "rules-status": "eslint-index lib/config/all.js  --format table",
     "rules-omitted": "eslint-index lib/config/all.js  --status omitted",
     "lint": "eslint . --max-warnings 0",
+    "prettier-check": "eslint --print-config .eslintrc.js | eslint-config-prettier-check",
```
ismail-syed: `.eslintrc.js` should point to ~`esnext`~ `prettier` config instead.

lib/config/jquery.js
```diff
@@ -3,9 +3,7 @@ module.exports = {
     jquery: true,
   },
 
-  plugins: [
-    'shopify',
-  ],
```
lemonmade, line 8: Are these changes the result of running prettier?

lib/config/prettier.js
```diff
@@ -0,0 +1,20 @@
...
+      {
+        singleQuote: true,
+        trailingComma: 'es5',
+        bracketSpacing: false,
+        jsxBracketSameLine: false,
+      },
+    ],
+
+    // rules to disable to prefer prettier
+    'shopify/binary-assignment-parens': 'off',
```
lemonmade, line 18: Can you explain this more

lib/config/prettier.js
```diff
@@ -0,0 +1,20 @@
...
+  extends: ['plugin:shopify/esnext', 'prettier'],
+
+  plugins: ['prettier'],
+
+  rules: {
+    'prettier/prettier': [
+      'error',
+      {
+        singleQuote: true,
+        trailingComma: 'es5',
```
lemonmade, line 11: Hmm, I think we need to change this for the ESNext config — any way of making this dynamic based on other configs passed in?

lib/rules/sinon-prefer-meaningful-assertions.js
lemonmade: Space before the last `*`

tests/.eslintrc.js
lemonmade: I like the alternative better

lib/config/flow.js
```diff
@@ -12,12 +12,7 @@ module.exports = {
     sourceType: 'module',
   },
 
-  plugins: [
-    'flowtype',
-    'shopify',
-  ],
+  plugins: ['flowtype', 'shopify'],
```
lemonmade, line 15: There's also a config for `prettier/flowtype` (in addition to one for React) — do we want those?

ismail-syed: Yup. That work can be in a future PR. We'll have to manually find out which rules from those configs we'll need to turn off.

lib/config/jquery.js
ismail-syed: Yup, it's within the 80 column size .

lib/config/prettier.js
ismail-syed: The prettier config will [disable](https://github.com/prettier/eslint-config-prettier/blob/master/index.js#L22-L24) the rule(s) regardless of the config passed in.

Extending this prettier config forces you to opt-in to the opinionated prettier experience. If a consumer wants, they can update the `prettier/prettier` rules in their config. 

```
    'prettier/prettier': [
      'error',
      {
        singleQuote: true,
        trailingComma: '<none|es5|all>',
        bracketSpacing: false,
        jsxBracketSameLine: false,
      },
    ],
```

lib/config/prettier.js
ismail-syed: `shopify/binary-assignment-parens` expects something like this: 

```
const always = (config === 'always');
```

This will cause prettier to complain:
```
$ eslint . --max-warnings 0

/Users/ismailsyed/src/github.com/Shopify/eslint-plugin-shopify/lib/rules/class-property-semi.js
  14:20  error  Replace `(config·===·'always')` with `config·===·'always'`  prettier/prettier
```

Unfortunately, we'll have to turn this rule off in favor of the 'prettier' experience. 
No pun intended, lol.

ismail-syed:
Comments addressed, ready for another 👀

GoodForOneFare (commented):
It's a clever idea to show the effect of Prettier in this PR, but it's making it difficult to sift out what has really changed.

From my first sift through, two things that I'm not comfortable with:
* Arrays being translated to single lines based on line width
  * This seems at odds with our "trailing comma makes it easier to track the originator of a change" philosophy
* `[{` being rewritten to have the opening brace on a new line

The second thing I'm sure I'll get over.  The first seems like it should be up to the developer.

ismail-syed:
I kept all the automated `--fix` with the prettier changes in this one commit, https://github.com/Shopify/eslint-plugin-shopify/pull/46/commits/5c2baf34ea7382494150a20e26504c0a9ce4cd43. I should have mentioned that in the PR description, sorry.
It might be helpful to view it in the individual commit instead.

package.json
```diff
@@ -49,6 +51,7 @@
     "isparta": "^4.0.0",
     "istanbul": "^0.4.5",
     "mocha": "^3.4.2",
+    "prettier": "^1.7.2",
```
ismail-syed, line 54: Move to `dependencies`

lemonmade: Why?

ismail-syed: So consumers don't have to install `prettier`.

lemonmade: I just worry about it needing to be aligned across repos, technically they would need to install it now in order to run the command alongside the linter

ismail-syed: Spoke offline with Chris. We're going to add prettier as an optional dependency since prettier is only required when extending the prettier config.

lemonmade (approved):
This feels fine to me, I'm OK with the occasional non-ideal formatting.

GoodForOneFare (approved):

PR #48, opened 2017-10-23
Replace all warns with errors

ismail-syed:
We pretty much treat all `warn` as `error` anyways, let's make this explicit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,9 +3,11 @@
 ## [Unreleased]
 
 ### Changed
+* Replace all `warn` with `error`
 * `space-before-function-paren` now uses `asyncArrow` option (eg. `async () => {}`)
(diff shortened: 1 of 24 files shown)


gf3 (approved):

GoodForOneFare:
Context:
```
*Chris*
the idea was that all lint errors were errors

we only made some warnings because errors were obnoxious in Atom at one point :stuck_out_tongue:

if we let warnings through then we a) need to have a harder rule on what is a warning and what is an error, and b) have some criteria for ignorable warnings

both seem really hard to do well so I would rather just follow the stylelint approach here (all things are errors)

if we have linting warnings we think are usuablly ignorable we should turn the rule off

*Tyler*
can we reclassify everything in our linting config as errors then?

if you’re running eslint outside the context of sewing-kit but are using the config then you have to explicitly set `max-warnings` on your lint task

which is a little obscure

*Chris*
I’d be happy with that
```

PR #53, opened 2017-11-29
Pull out esnext config from prettier config

ismail-syed:
This PR removes the extended `esnext` config from the `prettier` config. 

**Why?:**
This allows us to extend the `prettier` config on top of `typescript`/`typescript-react` (WIP) config without conflicts. 

This will be a breaking change; consumers will have to prepend the `esnext` config to their eslint config wherever the `prettier` eslint config is used.

Open to alternatives suggestions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,19 @@
 
 ## [Unreleased]
 
+### Changed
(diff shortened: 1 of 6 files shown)


lemonmade (approved):
This is fine, it's how everything other than the react config works already, and the react config should also work this way. 👍

PR #42, opened 2017-09-20
Expose a typescript & typescriptReact config

ismail-syed:
* This PR adds a consumable `typescript` & `typescriptReact` config for projects with `.ts` and `.tsx` files. 
* It's likely that some minor rules may have been missed, but I've disabled all the major one that conflicts with the ts parser. The parser is sometimes played nice and lists out which rule it failed on, other times it was all divide and conquer. 

Next steps: 
* Add `prettier`  plugins
* Update `lint` commands in `sewing-kit`. 
* Add `format` command in `sewing-kit`

Tophat instructions:

**NOTE:** The branches below are using beta releases of `eslint-plugin-shopify` with the work in this branch. 

---

### 🎩  in `sewing-kit`:
1. In `sewing-kit` checkout [`test-typescript-eslint-config`](https://github.com/Shopify/sewing-kit/commits/test-typescript-eslint-config) and run 
```bash
yarn install
```
2. `yarn run lint` to verify that it's linting as per the `typescript` config in this PR.

### 🎩 in `web`:

1. In `web` checkout [`test-typescript-eslint-config`](https://github.com/Shopify/web/commits/test-typescript-eslint-config) and run
```bash
yarn install
``` 

2. `yarn run lint` to verify that it's linting as per the `typescript` & `typescriptReact` config in this PR.

PR #54, opened 2017-12-01
Adding a typescript & typescript-react config

ismail-syed:
This PR adds a consumable typescript & typescriptReact config for projects with .ts and .tsx files.
The application of the `typescript` config is demonstrated in sewing-kit in [here](https://github.com/Shopify/sewing-kit/pull/438)

This is a revive of https://github.com/Shopify/eslint-plugin-shopify/pull/42


lib/config/rules/typescript.js
```diff
@@ -0,0 +1,110 @@
...
+  'no-dupe-class-members': 'off',
+
+  // flags typedef files with multiple modules with export default
+  'import/export': 'off',
+
+  // breaks typescript-eslint-parser
+  strict: 'off',
+  'shopify/prefer-early-return': 'off',
+  'array-callback-return': 'off',
+  'getter-return': 'off',
```
ismail-syed, line 109: This config still uses the stock `typescript-eslint-parser`. The workaround attempts to address these 4 rules revealed more concerns with the parser. 

The performance improvements of using eslint vs tslint easily outweigh loosing these 4 rules.

lib/config/rules/typescript-react.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
+  // Disable JS specific rules
+  'react/jsx-filename-extension': 'off',
+  'react/default-props-match-prop-types': 'off',
+
+  // Breaks typescript-eslint-parser
+  'react/jsx-indent': 'off',
```
ismail-syed, line 7: It sucks that the fixers for these rules break the parser, but using the prettier config with the parser option set to typescript gives us the features back.

lemonmade (approved):

PR #55, opened 2017-12-01
Add trailing comma function support to comma-dangle and prettier config

ismail-syed:
* `comma-dangle` will now enforce trailing commas in multi-line function calls via the `esnext` config. This was previously enforced through our babel config ([`babel/func-params-comma-dangle`](https://github.com/babel/eslint-plugin-babel/releases/tag/v4.0.0)) but the rule has been deprecated. We never updated the eslint `comma-dangle` to reflect the deprecation.

* the `trailingComma` setting in our `prettier` config has also been updated the `trailingComma` option to `all`. This will also enforce trailing commas in multiline function calls

* Added a `.prettierrc` file to the project so maintainers can get prettier formatting via editor plugins

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,5 @@
+{
+  "singleQuote": true,
+  "bracketSpacing": false,
+  "trailingComma": "all"
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 
 ### Changed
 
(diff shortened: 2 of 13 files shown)


ismail-syed:
@lemonmade, any ideas why CI is failing? 
Everything is fine locally, I tried rebuilding without cache too. 

**Update:** had to bump node in CI.

PR #61, opened 2017-12-21
Update codebase to use es5 trailingComma

ismail-syed:
* Format project codebase to use `es5` style for trailing comma so we don't break projects that are on node version 6 😥

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,5 +1,5 @@
 {
   "singleQuote": true,
   "bracketSpacing": false,
-  "trailingComma": "all"
+  "trailingComma": "es5"
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,9 @@
 # Changelog
 
-<!-- ## Unreleased -->
+## Unreleased
 
+### Changed
+* Changed `eslint-config-shopify` codebase to follow es5 trailingComma [[#61](https://github.com/Shopify/eslint-plugin-shopify/pull/61)]
 
 ## [18.3.0] - 2017-12-18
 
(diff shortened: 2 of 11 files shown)


lemonmade (approved):

kaelig:
The `comma-dangle` rule is still throwing errors in projects that don't use Prettier (seen in `polaris-react` when upgrading from v17 to v19-beta). Can we also fix this issue?

lemonmade:
What errors are you seeing? This fix was about this package itself not using trailing commas because it has to run in node, Polaris has a build step where they are stripped out so it is supposed to throw errors if there are no trailing commas in the source.

ismail-syed:
Do you have a branch I can look at?

kaelig:
Yes, you can have a look over there: https://github.com/Shopify/polaris-react/pull/885

lemonmade:
@kaelig you'll have to add a separate .eslintrc.js/ use glob patterns in your eslintrc to specify different rules for non-source files if they can't be run as-is when following the rule, it's not something we are really optimizing for given that we mostly just have to lint files that get compiled in some way.

kaelig:
Thanks, that's starting to make sense.

It's quite a bit of added cognitive load, could we align on one single setting for a more consistent DX?

Here are a few solutions I can think of:
- keep `'comma-dangle': 'all'`, and stop supporting Node 6 (version 8 is LTS after all)
- keep `'comma-dangle': 'all'`, and add a prepublish script to transpile the npm package down to es5
- revert the rule to follow `'comma-dangle': 'es5'` everywhere

ismail-syed:
> keep 'comma-dangle': 'all', and stop supporting Node 6 (version 8 is LTS after all)

I prefer this, that way we force the org to update to node 8.

lemonmade:
I am still not sure I understand. This is a project-level configuration issue, nothing to do with our linting config. We specify different configs for different environments, it's up to the user to choose the right one for them. Moving to Node 8 simplifies this because no project has to choose between comma dangle or not, but there will be other issues for which the answer is the same.

PR #63, opened 2018-01-03
Upgrade Prettier to ^1.9.2  (arrowParens: always) 🎉 & update dependencies

kaelig:
**⚠️this is likely to be a breaking change**

* Updated dependencies to their latest versions
* Updated prettier to 1.9.2, introducing a change in function parens style (set to `arrowParens: 'always'`):

    ```js
    // Before
    const foo = myArray.map(foo => {});

    // After
    const foo = myArray.map((foo) => {});
    ```

    #### ⚠️ Upgrade path

    Your project config files (`package.json`, `.prettierrc`, `.eslintrc`…)
    may need to be updated like so:

    ```diff
       "singleQuote": true,
       "bracketSpacing": false,
       "trailingComma": "es5",
    +  "arrowParens": "always"
    ```

* Added a `prettier` script: `yarn prettier` now prettifies source files
* Prettified source files using the new config
  
---

### Notable upgrades

- **eslint-config-prettier:** [`v2.6.0...v2.9.0`](https://github.com/prettier/eslint-config-prettier/compare/v2.6.0...v2.9.0)
  - Added: The [implicit-arrow-linebreak] rule
  - Added: The [react/jsx-one-expression-per-line] rule.
  - Added: The [lines-around-comment] rule (as a [special
    rule][lines-around-comment-special]).
  - Added: The [no-unexpected-multiline] rule (as a [special
    rule][no-unexpected-multiline-special]).
- **eslint-plugin-ava:** [`v4.2.2...v4.4.0`](https://github.com/avajs/eslint-plugin-ava/compare/v4.2.2...v4.4.0)
  - no notable rule changes
- **eslint-plugin-flowtype:** [`v2.39.1...v2.41.0`](https://github.com/gajus/eslint-plugin-flowtype/compare/v2.39.1...v2.41.0)
  - Added: [flowtype/no-flow-fix-me-comments](https://github.com/gajus/eslint-plugin-flowtype/blob/677e55c6a0f1dd355268a0f19618cd2696424c53/.README/rules/no-flow-fix-me-comments.md)
- **eslint-plugin-jsx-a11y:** [`v6.0.2...v6.0.3`](https://github.com/evcohen/eslint-plugin-jsx-a11y/compare/v6.0.2...v6.0.3)
  - no notable rule changes
- **eslint-plugin-react:** [`v7.4.0...v7.5.1`](https://github.com/yannickcr/eslint-plugin-react/compare/v7.4.0...v7.5.1)
  * Added: [`jsx-one-expression-per-line`][]
  * Added: [`destructuring-assignment`][]
  * Added: [`no-access-state-in-setstate`][]
  * Added: [`button-has-type`][]
  * Added: [`jsx-curly-brace-presence`][] rule ([#1310][] @jackyho112)
- **eslint-plugin-prettier:** [`v2.3.1...v2.4.0`](https://github.com/prettier/eslint-plugin-prettier/compare/v2.3.1...v2.4.0)
  - no notable rule changes
- **eslint-plugin-sort-class-members:** [`v1.2.2...v1.3.0`](https://github.com/bryanrsmith/eslint-plugin-sort-class-members/compare/v1.2.2...v1.3.0)
  - no notable rule changes
- **eslint-plugin-typescript:** [`v0.7.0...v0.8.1`](https://github.com/nzakas/eslint-plugin-typescript/compare/f860b4bf3b43e6129fc5ca28564b8ef75a79df5c...bbfb87860c5c6834403e9c08934ff41134118099)
  - Added: [`typescript/member-naming`](https://github.com/nzakas/eslint-plugin-typescript/tree/master/docs/rules/member-naming.md) — Enforces naming conventions for class members by visibility.
  - Added: [`typescript/no-array-constructor`](https://github.com/nzakas/eslint-plugin-typescript/tree/master/docs/rules/no-array-constructor.md) — Disallow generic `Array` constructors
- **typescript-eslint-parser:** [`v9.0.0...v11.0.0`](https://github.com/eslint/typescript-eslint-parser/compare/v9.0.0...v11.0.0)
  - no notable rule changes

[implicit-arrow-linebreak]: https://eslint.org/docs/rules/implicit-arrow-linebreak
[lines-around-comment-special]: https://github.com/prettier/eslint-config-prettier/blob/5399175c37466747aae9d407021dffec2c169c8b/README.md#lines-around-comment
[lines-around-comment]: https://eslint.org/docs/rules/lines-around-comment
[no-unexpected-multiline-special]: https://github.com/prettier/eslint-config-prettier/blob/5399175c37466747aae9d407021dffec2c169c8b/README.md#no-unexpected-multiline
[no-unexpected-multiline]: https://eslint.org/docs/rules/no-unexpected-multiline
[react/jsx-one-expression-per-line]: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-one-expression-per-line.md
[`jsx-one-expression-per-line`]: https://github.com/yannickcr/eslint-plugin-react/tree/master/docs/rules/jsx-one-expression-per-line.md
[`destructuring-assignment`]: https://github.com/yannickcr/eslint-plugin-react/tree/master/docs/rules/destructuring-assignment.md
[`no-access-state-in-setstate`]: https://github.com/yannickcr/eslint-plugin-react/tree/master/docs/rules/no-access-state-in-setstate.md
[`button-has-type`]: https://github.com/yannickcr/eslint-plugin-react/tree/master/docs/rules/button-has-type.md
[`jsx-curly-brace-presence`]: https://github.com/yannickcr/eslint-plugin-react/tree/master/docs/rules/jsx-curly-brace-presence.md

---

## Todo

- [x] Document all notable rule changes
- [x] Add an explicit 'off' setting for all new rules
- [x] Add a setting for a rule we missed in a previous upgrade: `eslint-plugin-import@2.8.0` [`exports-last`](https://github.com/benmosher/eslint-plugin-import/blob/master/CHANGELOG.md#added).


kaelig:
I'm not sure how to write tests, or if it's necessary at all to write any for the `arrowParens: 'always'` case. @ismail-syed thoughts?

package.json
```diff
@@ -48,49 +49,50 @@
...
     "eslint-plugin-babel": "^4.1.2",
     "eslint-plugin-chai-expect": "^1.1.1",
-    "eslint-plugin-flowtype": "^2.32.1",
-    "eslint-plugin-import": "^2.7.0",
-    "eslint-plugin-jsx-a11y": "^6.0.2",
-    "eslint-plugin-lodash": "^2.4.4",
+    "eslint-plugin-flowtype": "^2.40.1",
+    "eslint-plugin-import": "^2.8.0",
+    "eslint-plugin-jsx-a11y": "^6.0.3",
+    "eslint-plugin-lodash": "^2.5.0",
```
GoodForOneFare: It looks like dependencies are just being bumped without updating our configs to reflect new rules / options.  Are we okay with just using whatever the defaults are?

e.g.,

`eslint-plugin-lodash@2.5.0` [introduces `import-scope`](https://github.com/wix/eslint-plugin-lodash/blob/master/CHANGELOG.md#added-1).  

`eslint-plugin-import@2.8.0` [introduces `exports-last`](https://github.com/benmosher/eslint-plugin-import/blob/master/CHANGELOG.md#added).

kaelig: Thanks for asking!

> eslint-plugin-lodash@2.5.0 introduces import-scope.

Looks like we already had a rule for that, maybe the yarn version was already set to 2.5.0.

```js
  // Prefer a specific import scope (e.g. lodash/map vs lodash)
  'lodash/import-scope': ['error', 'method'],
```

As for `exports-last`, it's disabled by default. Should I explicitly set it to `'off'`?

GoodForOneFare: I believe that we explicitly set a value for all options.  @lemonmade, that's correct?

Most likely, there are many more new rules + new options.  I'd rather someone trawl through changelogs and produce curated PRs, instead of a shotgun update.

lemonmade: Yes, we try to provide explicit settings for all rules.

kaelig: > Most likely, there are many more new rules + new options. I'd rather someone trawl through changelogs and produce curated PRs, instead of a shotgun update.

I checked yarn.lock and it turns out we were already using the latest version of those 2 plugins. I guess the trick is to control which version is in the yarn.lock and make sure to document those upgrades too.

Let me go through the diffs and upgrade this PR.

kaelig: I've updated the body of the PR with a list of all new rules and a TODO. I'll add explicit settings for all those rules as well.

kaelig: @GoodForOneFare @lemonmade I've just updated the PR with new rules explicitly set to `off` when applicable. Can you please have another look?

lib/config/rules/typescript.js
```diff
@@ -10,15 +10,21 @@ module.exports = {
...
+  // enforces naming conventions for class members by visibility.
+  'typescript/member-naming': 'off',
+
   // enforces /// <reference /> is not used. (no-reference from TSLint)
   'typescript/no-triple-slash-reference': 'error',
 
-  // enforces the any type is not used. (no-any from TSLint)
-  'typescript/no-explicit-any': 'off',
+  // disallow generic Array constructors
+  'typescript/no-array-constructor': 'error',
```
kaelig, line 20: This is set to 'error' (and not turned off), to replicate this ESLint rule:

https://github.com/Shopify/eslint-plugin-shopify/blob/9c33d006cb34d20825d6fab5331ba04d050071e9/lib/config/rules/stylistic-issues.js#L85-L86

ESLint original rule: https://eslint.org/docs/rules/no-array-constructor
eslint-plugin-typescript: https://github.com/nzakas/eslint-plugin-typescript/blob/master/docs/rules/no-array-constructor.md

kaelig:
I think I'm done with this PR - can you please give it another 👀? Merci!

GoodForOneFare (changes requested):
At this point, it would be better to split this into two PRs so that the Prettier upgrade can ship today.

Are most of the rules turned off because they're not useful, or to expedite the PR?

At minimum, the PR is still missing:
* lines-between-class-members
* multiline-comment-style
* flowtype/no-mutable-array
* flowtype/no-unused-expressions
* promise/no-return-in-finally

package.json
```diff
@@ -48,50 +49,51 @@
...
+    "prettier": "^1.9.2",
+    "react": "^16.2.0",
+    "react-dom": "^16.2.0"
   },
   "peerDependencies": {
     "eslint": "<5 >=4.7.1"
   },
   "optionalDependencies": {
-    "prettier": "<2.0 >= 1.7.2"
+    "prettier": "<2.0 >= 1.9.2"
```
GoodForOneFare, line 77: I dislike that this installs `prettier` even when consuming projects don't use it.  I'd like to move it to `peerDependencies`.

Make sense, @ismail-syed and @lemonmade?

kaelig: Good point, but is this in the scope of this PR?

lemonmade: I thought optional dependencies was the right thing — doesn't `peer` require the installation, even if unused?

kaelig:
You raise good points, and I'm sorta concerned by the overhead of maintaining this growing number of rules 🧐

As far as this PR is concerned, the missing rules are unrelated (they can be added separately, no?). Should we really hold off merging it?

package.json
ismail-syed: > I dislike that this installs prettier even when consuming projects don't use it

I agree. Optional dependencies aren't what we expected them to be. 

As per yarn [docs](https://yarnpkg.com/lang/en/docs/dependency-types/#toc-optionaldependencies)
> Optional dependencies are just that: optional. If they fail to install, Yarn will still say the install process was successful.

prettier will still be installed; in the chance that it fails, `yarn install` will still succeed. 

I suggest we change it to a peer dependency like we do with `eslint`.

GoodForOneFare: Here's a repo that confirms the optional/peer behaviour: https://github.com/GoodForOneFare/dependency-install-types

* Optionals are installed for all consumers
* Peers generate warnings during `yarn install`, and are not installed

GoodForOneFare:
If rules were added as `off` to expedite this PR, they should be removed, or at least flagged with `TODO`.

Since the last dependency bump PR serves as a good reference for the next PR, I'd rather produce a well-curated example for everyone.  That said, this isn't a formally documented rule, so ¯\\_(ツ)_/¯

I'll defer to the original assigned reviewers, and bow out of this discussion.

package.json
kaelig: That sounds like something that should be discussed outside of this PR, unless I’m missing something?

lib/config/rules/react.js
```diff
@@ -3,10 +3,14 @@ module.exports = {
 
   // Enforces consistent naming for boolean props
   'react/boolean-prop-naming': 'off',
+  // Prevent usage of button elements without an explicit type attribute
+  'react/button-has-type': 'off',
```
lemonmade: `error`

lib/config/rules/react.js
```diff
@@ -15,6 +19,8 @@ module.exports = {
   'react/forbid-foreign-prop-types': 'error',
   // Forbid certain propTypes
   'react/forbid-prop-types': ['error', {forbid: ['any', 'array']}],
+  // Prevent using this.state within a this.setState
+  'react/no-access-state-in-setstate': 'off',
```
lemonmade: `error`

lib/config/rules/react.js
```diff
@@ -90,6 +96,8 @@ module.exports = {
   'react/jsx-closing-bracket-location': ['error', {location: 'tag-aligned'}],
   // Validate closing tag location in JSX
   'react/jsx-closing-tag-location': 'error',
+  // Enforce curly braces or disallow unnecessary curly braces in JSX props and/or children
+  'react/jsx-curly-brace-presence': 'off',
```
lemonmade: [`'error', 'never']`

lemonmade (approved):
I left what I think are the right configuration options for rules we actually want to turn on. I am fine with changing those in a follow-up PR and to ship this as-is (although in future, I'd recommend only upgrading the things you actually need to avoid having this kind of creep in the first place)

lib/config/rules/typescript.js
ismail-syed: Can we add this to the CHANGELOG, please.

CHANGELOG.md
```diff
@@ -11,8 +11,34 @@
...
+
+    #### ⚠️ Upgrade path
+
+    Your project config files (`package.json`, `.prettierrc`, `.eslintrc`…)
+    may need to be updated like so:
+
+    ```diff
+       "singleQuote": true,
+       "bracketSpacing": false,
+       "trailingComma": "es5",
```
ismail-syed, line 69: Our config uses `trailingComma: 'all'`

ismail-syed (approved):
Looks good to me. Just a couple comments. 

* Can we append `& update dependencies` to the title of the PR. When this gets merged it'll be easier to read through the commit history and understand the changes. 

* Can we add the newly introduced rules to the changelog as well - [example](https://github.com/Shopify/eslint-plugin-shopify/blob/d9d4a5fd42c64435f4dfda3a73d41c1ed13fdc95/CHANGELOG.md#1710---2017-09-19)

ismail-syed:
Did we do a beta test with this on any repos?

CHANGELOG.md
kaelig: Ha, is that something we should also standardize? Looks like not all codebases use it like that.

lemonmade: Pretty sure it is consistent, our listing rules have always enforced trailing commas.

lib/config/rules/react.js
kaelig: 👍

lib/config/rules/react.js
kaelig: 👍

lib/config/rules/react.js
kaelig: 👍

kaelig:
> Can we append & update dependencies to the title of the PR. When this gets merged it'll be easier to read through the commit history and understand the changes.

👍 

> Can we add the newly introduced rules to the changelog as well

👍 

* Added [`implicit-arrow-linebreak`][] rule
* Added [`lines-around-comment`][] rule (as a [special
  rule][lines-around-comment-special]).
* Added [`no-unexpected-multiline`][] rule (as a [special rule][no-unexpected-multiline-special]).
* Added [`flowtype/no-flow-fix-me-comments`](https://github.com/gajus/eslint-plugin-flowtype/blob/677e55c6a0f1dd355268a0f19618cd2696424c53/.README/rules/no-flow-fix-me-comments.md)
* Added [`react/jsx-one-expression-per-line`][]
* Added [`react/destructuring-assignment`][]
* Added [`react/no-access-state-in-setstate`][]
* Added [`react/button-has-type`][]
* Added [`react/jsx-curly-brace-presence`][]
* Added [`typescript/member-naming`](https://github.com/nzakas/eslint-plugin-typescript/tree/master/docs/rules/member-naming.md)
* Added [`typescript/no-array-constructor`](https://github.com/nzakas/eslint-plugin-typescript/tree/master/docs/rules/no-array-constructor.md)

[`implicit-arrow-linebreak`]: https://eslint.org/docs/rules/implicit-arrow-linebreak
[lines-around-comment-special]: https://github.com/prettier/eslint-config-prettier/blob/5399175c37466747aae9d407021dffec2c169c8b/README.md#lines-around-comment
[`lines-around-comment`]: https://eslint.org/docs/rules/lines-around-comment
[no-unexpected-multiline-special]: https://github.com/prettier/eslint-config-prettier/blob/5399175c37466747aae9d407021dffec2c169c8b/README.md#no-unexpected-multiline
[`no-unexpected-multiline`]: https://eslint.org/docs/rules/no-unexpected-multiline
[`react/jsx-one-expression-per-line`]: https://github.com/yannickcr/eslint-plugin-react/tree/master/docs/rules/jsx-one-expression-per-line.md
[`react/destructuring-assignment`]: https://github.com/yannickcr/eslint-plugin-react/tree/master/docs/rules/destructuring-assignment.md
[`react/no-access-state-in-setstate`]: https://github.com/yannickcr/eslint-plugin-react/tree/master/docs/rules/no-access-state-in-setstate.md
[`react/button-has-type`]: https://github.com/yannickcr/eslint-plugin-react/tree/master/docs/rules/button-has-type.md
[`react/jsx-curly-brace-presence`]: https://github.com/yannickcr/eslint-plugin-react/tree/master/docs/rules/jsx-curly-brace-presence.md

> Did we do a beta test with this on any repos?

I don't think we have. What codebase would be a good candidate for a test?

CHANGELOG.md
ismail-syed: Which code bases don't use it? They should be, unless they have a good reason not too. 

In the case of `eslint-plugin-shopify`, the repo itself uses es5 so we don't break consumers who may be on older versions of node. (eg. u2). We could add a build step in this repo to solve that, but don't think it's worth the effort.

ismail-syed:
> I don't think we have. What codebase would be a good candidate for a test?

I usually do `web` and `sewing-kit`. Sometimes admin to verify if there's no major 💥 

However, it might be nice to spread out the testing codebases. Anything blocking `polaris-react` or `polaris-styleguide` from being candidates.

kaelig:
Thanks! Can you do `web` and `sewing-kit`? I'll do `polaris-react`.

On `polaris-styleguide` we are so many versions of sewing-kit behind that it will definitely 💥

PR #67, opened 2018-01-07
Web linting rules

lemonmade:
Fixes https://github.com/Shopify/sewing-kit/issues/461
Fixes https://github.com/Shopify/web/issues/1752
Fixes https://github.com/Shopify/web/issues/162
Fixes https://github.com/Shopify/web/issues/118

Adds a bunch of linting rules we felt would be good to have from our work on web. It's a really big PR, which I can break up if people really want, but a lot of it is just docs and tests.


docs/rules/jsx-no-complex-expressions.md
```diff
@@ -0,0 +1,26 @@
+# Disallows complex expressions embedded in in JSX. (jsx-no-complex-expressions)
+
+Having complex conditionals embedded in JSX can make it harderand less HTML-like to read. This rule aims to improve the readability of JSX by preventing the use of complex expressions.
```
GoodForOneFare: `harderand` => `harder and`.  Even with that fixed, it's still oddly worded.  The "less HTML-like" is qualifying "harder".  

"Less idiomatic" covers both, and sounds super smart 🤓

README.md
```diff
@@ -114,10 +115,16 @@ This plugin provides the following custom rules, which are included as appropria
...
+- [jsx-no-hardcoded-content](docs/rules/jsx-no-hardcoded-content.md): Disallows hardcoded content in JSX.
 - [no-useless-computed-properties](docs/rules/no-useless-computed-properties.md): Prevents the usage of unnecessary computed properties.
+- [polaris-no-bare-stack-item](docs/rules/polaris-no-bare-stack-item.md): Disallow the use of Polaris’s `Stack.Item` without any custom props.
+- [polaris-prefer-sectioned-prop](docs/rules/polaris-prefer-sectioned-prop.md): Prefer the use of the `sectioned` props in Polaris components instead of wrapping all contents in a `Section` component.
 - [prefer-class-properties](docs/rules/prefer-class-properties.md): Prefer class properties to assignment of literals in constructors.
 - [prefer-early-return](docs/rules/prefer-early-return.md): Prefer early returns over full-body conditional wrapping in function declarations.
 - [prefer-twine](docs/rules/prefer-twine.md): Prefer Twine over Bindings as the name for twine imports.
+- [react-initialize-state](docs/rules/react-initialize-state.md): Requires that React component state be initialize when it has a non-empty type.
```
GoodForOneFare: `be initialize` => `be initialized`

docs/rules/jsx-no-hardcoded-content.md
```diff
@@ -0,0 +1,142 @@
+# Disallows hardcoded content in JSX. (jsx-no-hardcoded-content)
+
+Many JSX components accept children and other properties that are intended for UI content. In an internationalized app, you generally provide content to these components through a translation function, not directly as strings. This rule allows you to enforce that components not accept harcoded literal content.
```
GoodForOneFare: General nit that I won't belabour: the use of "you" in this PR leads to padded, repetitious sentences.  Not 100% of them, but enough.

docs/rules/react-initialize-state.md
```diff
@@ -0,0 +1,33 @@
+# Requires that React component state be initialize when it has a non-empty type (react-initialize-state)
```
GoodForOneFare: `be initialize` => `be initialized`

lib/rules/react-initialize-state.js
```diff
@@ -0,0 +1,81 @@
+const Components = require('eslint-plugin-react/lib/util/Components');
+const {uncast, getName} = require('../utilities');
+
+module.exports = {
+  meta: {
+    docs: {
+      description:
+        'Requires that React component state be initialize when it has a non-empty type.',
```
GoodForOneFare: `be initialize` => `be initialized`

docs/rules/jsx-no-hardcoded-content.md
```diff
@@ -0,0 +1,142 @@
...
+
+### `allowStrings`
+
+Allows hardcoded string and template literals as content. Optional, defaults to `false`.
+
+With `{allowStrings: false}`, the following patterns are considered warnings:
+
+```js
+<div>Content</div>
+<MyComponent>{someContent()}{` and ${moreContent()}`}</MyComponent>
```
GoodForOneFare: Not clear on what's being flagged here.  Is it:
1. The template string?
2. Two translation function calls making a single string?

If it's both, I'd prefer to have separate examples.  If not, I'd rather have a simpler example.

docs/rules/jsx-no-hardcoded-content.md
```diff
@@ -0,0 +1,142 @@
...
+```
+
+### `checkProps`
+
+Specifies a list of props to check for hardcoded content, according to the rules set in `allowStrings` and `allowNumbers`. Note that this check will not be able to catch hardcoded content that was not passed directly (for example, props that are set to variables initially, or props that are spread). Optional, defaults to an empty array.
+
+With `{checkProps: ['foo'], allowStrings: false}`, the following patterns are considered warnings:
+
+```js
+<MyComponent>Content</MyComponent>
```
GoodForOneFare, line 53: I get why this would be flagged, but how does it relate to `checkProps`?

docs/rules/jsx-no-hardcoded-content.md
```diff
@@ -0,0 +1,142 @@
...
+<MyComponent foo={42} />
+<MyComponent foo={content()} />
+<MyComponent bar="baz" />
+```
+
+### `modules`
+
+Allows you to specify custom validation logic for components in different modules. This option should be an object where:
+
+* The key is the name of the module. For node modules, simply use the name by which you would import it. For local modules, specify the root-relative path of the module (so, for example, `app/components` will set options for anything that imports from `app/components.js`/ `app/components/index.js`, regardless of the relative path used for the import itself). Note that this check will also respect any custom resolution logic used for imports specified for [eslint-plugin-import](https://github.com/benmosher/eslint-plugin-import).
```
GoodForOneFare: The first bullet point is a real mouthful.   Nested bullet points are ugly, but they'd be footholds for my chihuahua attention span.

-----

eslint-plugin-shopify 2018-01-08 10-34-40

docs/rules/jsx-no-hardcoded-content.md
```diff
@@ -0,0 +1,142 @@
...
+The following patterns are not warnings:
+
+```js
+<div>{someContent(42)}</div>
+<MyComponent amount={42} />
+```
+
+### `checkProps`
+
+Specifies a list of props to check for hardcoded content, according to the rules set in `allowStrings` and `allowNumbers`. Note that this check will not be able to catch hardcoded content that was not passed directly (for example, props that are set to variables initially, or props that are spread). Optional, defaults to an empty array.
```
GoodForOneFare, line 48: Would `excludeProps` lead to more coverage?

lib/rules/jsx-no-complex-expressions.js
```diff
@@ -0,0 +1,26 @@
+const CONDITIONAL_EXPRESSION_WARNING = [
+  'Don’t use conditional expressions inside JSX',
```
GoodForOneFare: No _you_!  This is going to be an annoying rule for many devs, so it'd be good to lead with why it exists.

lib/rules/jsx-no-hardcoded-content.js
```diff
@@ -0,0 +1,178 @@
...
+        }
+
+        const elementName = context
+          .getSourceCode()
+          .getText(node.openingElement.name);
+
+        if (check.prop === 'children') {
+          context.report(
+            node,
+            `You must not use hardcoded content as the children of the ${
```
GoodForOneFare: 👎 No _you_!  Give more _why_ context, and less _do what I say_ tone.

docs/rules/polaris-no-bare-stack-item.md
```diff
@@ -0,0 +1,27 @@
...
+<Polaris.Stack.Item>Content</Polaris.Stack.Item>
+<PolarisStack.Item>Content</PolarisStack.Item>
+```
+
+The following patterns are not warnings:
+
+```js
+import {Stack} from '@shopify/polaris';
+
+<Stack.Item fill>Content</Stack.Item>
```
GoodForOneFare, line 26: If dynamically imported components aren't covered, it might be worth including an example, or mentioning it in the rule description.
  
I'm on the fence about it, so ¯\\_(ツ)_/¯

lib/rules/react-initialize-state.js
```diff
@@ -0,0 +1,81 @@
...
+    let classInfo = null;
+
+    return {
+      ClassDeclaration(node) {
+        if (!utils.isES6Component(node)) {
+          return;
+        }
+
+        classInfo = {
+          hasStateType: classHasNonEmptytateType(node),
```
GoodForOneFare: `classHasNonEmptytateType` => `classHasNonEmptyStateType`

lib/rules/react-type-state.js
```diff
@@ -0,0 +1,48 @@
...
+          !inTypeScriptReactComponent ||
+          getName(node.key) !== 'state' ||
+          node.typeAnnotation != null
+        ) {
+          return;
+        }
+
+        context.report(
+          node,
+          'You must match the type of the state instance property to the second type parameter of your React component.'
```
GoodForOneFare: `match` => `add`.

GoodForOneFare (approved):
Really nicely written rules, and very solid tests 👍  Running this in admin adds ~1.5 seconds to `yarn run lint`, so it's a lot of extra coverage at minimal cost.

GoodForOneFare:
Ah, it looks like it adds 8-12 seconds of linting time to `web`.  Still worth it, IMO, unless it degrades VSCode's performance.

docs/rules/jsx-no-hardcoded-content.md
lemonmade: I generally prefer a whitelist over a blacklist, I imagine we will export this from Polaris at some point and so it can stay up-to-date regardless

docs/rules/jsx-no-hardcoded-content.md
lemonmade: I just want to demonstrate that children are still checked regardless

lib/rules/jsx-no-hardcoded-content.js
lemonmade: I don't think we can give a reason why, it will depend on the app how they should actually handle this one.

lemonmade:
@GoodForOneFare will ship this, we can monitor the impact on editor performance once it's in web. I imagine all the extra time is probably in the hardcoded content rule, if it is affecting performance I think there are a few (potentially slightly-unsafe) cacheing mechanisms we can put in.

PR #69, opened 2018-01-08
Add prefer-module-scope-constants rule

lemonmade:
A smaller new rule PR this time :P This one came up during a recent PR from the discounts team.

### DIFF
--- a/docs/rules/prefer-module-scope-constants.md
+++ b/docs/rules/prefer-module-scope-constants.md
@@ -29,7 +29,7 @@ The following patterns are not warnings:
 const MY_VALUE = true;
 
 function myFunction() {
-  const otherVlaue = false;
+  const otherValue = false;
 }
 ```
 
--- a/docs/rules/prefer-module-scope-constants.md
+++ b/docs/rules/prefer-module-scope-constants.md
@@ -1,0 +1,36 @@
+# Prefer that screaming snake case variables always be defined using `const`, and always appear at module scope. (prefer-module-scope-constants)
+
+We use screaming snake case variables to denote truly constant data used for other operations in the file. However, we don’t use this naming convention for all constants — only those that are only ever generated once, at module definition. This rule restricts this naming convention to be used only for these types of constants.
+
+## Rule details
+
+Two different errors are flagged by the rule: screaming snake case bindings that are not defined using `const`, and those that do not appear at module scope.
+
+The following patterns are considered warnings:
+
+```js
+// these kinds of values should never change
+let MY_VALUE = true;
+
+function myFunction() {
+  // this is just a regular `const`, use camelcase instead.
+  const OTHER_VALUE = false;
+}
+
+if (condition()) {
+  // same here, just a regular `const`.
+  const FINAL_VALUE = true;
+}
+```
+
+The following patterns are not warnings:
+
+```js
+const MY_VALUE = true;
+
+function myFunction() {
+  const otherVlaue = false;
+}
+```
+
+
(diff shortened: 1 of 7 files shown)


lib/rules/prefer-module-scope-constants.js
```diff
@@ -0,0 +1,47 @@
...
+            'You must use `const` when using screaming snake case variables. If this is not a constant, use camelcase instead.'
+          );
+          return;
+        }
+
+        const scope = context.getScope();
+        if (scope.type !== 'module') {
+          context.report(
+            node,
+            'You must placee screaming snake case at module scope. If this is not meant to be a module-scoped variable, use camelcase instead.'
```
kaelig: `placee` -> `place` ?

lib/rules/prefer-module-scope-constants.js
```diff
@@ -0,0 +1,47 @@
...
+        const {id} = node;
+
+        if (id.type !== 'Identifier' || id.name !== id.name.toUpperCase()) {
+          return;
+        }
+
+        if (!inConstDeclaration) {
+          context.report(
+            node,
+            'You must use `const` when using screaming snake case variables. If this is not a constant, use camelcase instead.'
```
kaelig: `camelcase` -> `camelCase`?

ismail-syed:
This will flag tests structured as follows, since the const isn't at the module scope.
 
```
describe('test', () => {
    const FOO = true;

    beforeEach(() => {
      console.log(FOO);
    });
});
```

Are these valid cases for ignoring the rule?

ismail-syed (approved):

ismail-syed:
Can we also update the CHANGELOG.md too.

docs/rules/prefer-module-scope-constants.md
```diff
@@ -0,0 +1,36 @@
...
+}
+```
+
+The following patterns are not warnings:
+
+```js
+const MY_VALUE = true;
+
+function myFunction() {
+  const otherVlaue = false;
```
GoodForOneFare: `Vlaue` => `Value`.

lib/rules/prefer-module-scope-constants.js
GoodForOneFare: `using` => `defining`.

tests/lib/rules/prefer-module-scope-constants.js
```diff
@@ -0,0 +1,64 @@
...
+
+ruleTester.run('prefer-module-scope-constants', rule, {
+  valid: [
+    {code: 'const FOO = true;', parserOptions},
+    {code: 'const foo = true;', parserOptions},
+    {code: '{ const foo = true; }', parserOptions},
+    {code: 'const foo = true, FOO = true;', parserOptions},
+    {code: 'const {FOO} = bar', parserOptions},
+    {code: '{ const {FOO} = bar; }', parserOptions},
+    {code: 'function foo() { const {FOO} = bar; }', parserOptions},
```
GoodForOneFare, line 30: Why does this pass?  Isn't `FOO` in `foo`'s scope?

GoodForOneFare (approved):
CHANGELOG update and 🚢

tests/lib/rules/prefer-module-scope-constants.js
lemonmade: It's destructured, which I think we want to permit in case you did a namespace import from a module and want to get some of its properties.

GoodForOneFare: Makes sense 👍

lib/rules/prefer-module-scope-constants.js
lemonmade: @kaelig I think `camelcase` is the correct way to spell it in a sentence

lemonmade:
@ismail-syed yes, the example you posted is something we'd want to discourage.

PR #71, opened 2018-01-09
Fix `rules-status` / `rules-omitted` scripts

GoodForOneFare:
Some problems:
- `all` was trying to include non-existent plugins 😬
- `all` was trying to require non-existent rule files

🎩 
* `yarn run rules-status` should produce a nicely formatted table
* `yarn run rules-omitted` should produce a nice list of rule names

### DIFF
--- a/lib/config/all.js
+++ b/lib/config/all.js
@@ -35,9 +35,6 @@ module.exports = {
     'shopify',
     'sort-class-members',
     'typescript',
-    'typescript-prettier',
-    'typescript-react',
-    'webpack',
   ],
 
   rules: merge(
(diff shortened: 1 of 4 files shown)


GoodForOneFare:
We could add a test that just runs `node lib/config/all`.  That'd fail CI on misconfigurations.  Thoughts?

ismail-syed:
Sounds like a good test to have 👍

ismail-syed (approved):

GoodForOneFare:
Added that test.  Not pretty, but it catches invalid plugins, and flags non-existent rule files.

Issue #75, opened 2018-01-15
shopify/react-initialize-state throws on valueless class properties


GoodForOneFare:
Fails at line 33's `node.value.type` check because value is undefined:
https://github.com/Shopify/eslint-plugin-shopify/blob/99115364ed40fdbd7d1a50e5747a8469963d7943/lib/rules/react-initialize-state.js#L29-L36

Sample crashy input:
```ts
class Foo extends React.PureComponent<ComposedProps, State> {
  state: State;

  constructor(props: ComposedProps) {
    super(props);

    this.state = {selectedSrc};
  }
}
```

cc @lemonmade

PR #76, opened 2018-01-16
Fix class instance member with no value

lemonmade:
Fixes https://github.com/Shopify/eslint-plugin-shopify/issues/75


GoodForOneFare (approved):
No issues running this in `web`, now 👍

PR #79, opened 2018-02-11
[DO NOT MERGE] Update typescript-eslint-parser for TypeScript 2.7 support

ismail-syed:
`typescript-eslint-parser` will fail on projects upgrading to Typescript 2.7. 
screen shot 2018-02-11 at 12 07 55 am

Merging this PR will address those concerns but introduce new ones. Old projects on 2.6.x will see the inverse of the screenshot above. If we don't want to enforce TypeScript 2.7.x, we can close this PR. 

For projects wanting to use TS 2.7, an acceptable workaround imo is to update `typescript-eslint-parser` to the latest in their respective projects `package.json` and run `yarn install`: 
```
  "resolutions": {
    "typescript-eslint-parser": "13.0.0"
  }
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,6 +97,6 @@
     "eslint-plugin-typescript": "^0.8.1",
     "merge": "^1.2.0",
     "pkg-dir": "^2.0.0",
-    "typescript-eslint-parser": "^11.0.0"
+    "typescript-eslint-parser": "^13.0.0"
   }
 }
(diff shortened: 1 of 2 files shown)


kaelig:
Should be addressed in #84 - IMHO it's okay to enforce the latest version of typescript as it follows our engineering values.

PR #84, opened 2018-03-13
Add support for ESLint doc url meta, update dependencies, support TypeScript 2.7

kaelig:
Released under version `20.0.0`.

---

* **Breaking:** the version of TypeScript supported by this plugin is 2.7.x (in line with [typescript-eslint-parser](https://github.com/eslint/typescript-eslint-parser)’s TypeScript support)
* Updated dependencies that support the new [ESLint documentation URL metadata](https://github.com/eslint/eslint/issues/6582). Errors from these plugins are accompanied by a link to the documentation for the broken rule.
* Updated various other dependencies to their latest versions
* Dependencies are now strictly versioned for tighter control over the exact rules the plugin enforces.

| Package | old | new |
| ------- | --- | --- |
| `eslint-plugin-ava` | `^4.4.0` | `4.5.1` |
| `eslint-plugin-import` | `^2.8.0` | `2.9.0` |
| `eslint-plugin-jest` | `^21.5.0` | `21.14.1` |
| `eslint-plugin-lodash` | `^2.5.0`  | `2.6.1` |
| `eslint-plugin-node` | `^5.2.1`  | `6.0.1` |
| `eslint-plugin-prettier` | `^2.4.0`  | `2.6.0` |
| `eslint-plugin-promise` | `^3.6.0`  | `3.7.0` |
| `eslint-plugin-react` | `^7.5.1`  | `7.7.0` |
| `eslint-plugin-typescript` | `^0.8.1`  | `0.10.0` |

* Added rules:
  - [`react/forbid-dom-props`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/forbid-dom-props.md) (disabled)
  - [`react/jsx-child-element-spacing`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-child-element-spacing.md) **error**
  - [`react/jsx-max-depth`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-max-depth.md) (disabled)
  - [`react/jsx-sort-default-props`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-sort-default-props.md) (disabled)
  - [`react/no-this-in-sfc`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-this-in-sfc.md) **error**
  - [`import/group-exports`](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/group-exports.md) (disabled)
  - [`import/no-self-import`](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-self-import.md) **error**
  - [`import/no-default-export`](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-default-export.md) (disabled)
  - [`import/no-useless-path-segments`](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-useless-path-segments.md) **error**
  - [`jest/prefer-expect-assertions`](https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/prefer-expect-assertions.md) (disabled)
  - [`jest/valid-expect-in-promise`](https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/valid-expect-in-promise.md) **error**
  - [`jest/valid-describe`](https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/valid-describe.md) **error**
  - [`jest/consistent-test-it`](https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/consistent-test-it.md) **error**
  - [`jest/no-test-prefixes`](https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/no-test-prefixes.md) **error**
  - [`jest/lowercase-name`](https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/lowercase-name.md) **error**
  - [`jest/no-jest-import`](https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/no-jest-import.md) **error**
  - [`promise/valid-params`](https://github.com/xjamundx/eslint-plugin-promise/blob/master/docs/rules/valid-params.md) **error**
  - [`promise/no-new-statics`](https://github.com/xjamundx/eslint-plugin-promise/blob/master/docs/rules/no-new-statics.md) (disabled)
  - [`typescript/explicit-function-return-type`](https://github.com/nzakas/eslint-plugin-typescript/blob/master/docs/rules/explicit-function-return-type.md) (disabled)
  - [`typescript/no-non-null-assertion`](https://github.com/nzakas/eslint-plugin-typescript/blob/master/docs/rules/no-non-null-assertion.md) **error**
* Updated `import/extensions` due to changes in its implementation: some extensions are explicitly allowed in `import`s: `.svg`, `.png`, `.jpg`, `.ico`. `.json` is still required as well.
* Chore: updated CircleCI from v1 to v2


kaelig:
## Request for comments

cc @lemonmade @ismail-syed @GoodForOneFare 

### `jest/lowercase-name`

I believe @lemonmade likes having tests that start with lowercase, so I turned on the new rule `jest/lowercase-name`, however it fails on this type of code:

> `describe`s should begin with lowercase

```js
describe('ModuleConcatenationPlugin', () => {
```

Should I turn `jest/lowercase-name` back off?

---

### `jest/valid-describe`

Fails on this code from sewing-kit:

> No async describe callback

```js
  describe('cache', async () => {
    it('always sets cacheing to true', async () => {
      expect(
        await webpackConfig(createWorkspace({env: client})),
      ).toHaveProperty('cache', true);
      expect(
        await webpackConfig(createWorkspace({env: server})),
      ).toHaveProperty('cache', true);
    });
  });
```

@GoodForOneFare any thoughts? Are async callbacks generally to be avoided or should we allow it?

---

### `jest/consistent-test-it`

Fails on this code from sewing-kit:

> Prefer using 'test' instead of 'it'

```js
it('fails', () => {
  expect(true).toBe(false);
});
```

By default, the rule enforces `test` at the root of test files, and `it` when nested under `describe`.

Thoughts?

cartogram:
Personally, I would prefer the `jest/lowercase-name` only be enforced on `it` and not `describe` (https://github.com/jest-community/eslint-plugin-jest/pull/92). My guess is that would remove most of the failures.

.circleci/config.yml
```diff
@@ -0,0 +1,21 @@
...
+            # when this file is changed, this key will fail
+            - v1-npm-deps-{{ arch }}-{{ checksum "yarn.lock" }}
+            # Find the most recent cache used from any branch
+            - v1-npm-deps-{{ arch }} # used if above checksum fails
+      - run: yarn
+      - save_cache:
+         key: v1-npm-deps-{{ arch }}-{{ checksum "yarn.lock" }}
+         paths:
+           - node_modules
+      - run: yarn run check
```
lemonmade, line 21: Probably would have been better as a separate PR?

lib/config/rules/import.js
```diff
@@ -61,7 +65,17 @@ module.exports = {
...
-  'import/extensions': ['error', {js: 'never', json: 'always'}],
+  'import/extensions': [
+    'error',
+    {
+      js: 'never',
+      json: 'always',
+      svg: 'always',
+      png: 'always',
+      jpg: 'always',
+      ico: 'always',
```
lemonmade, line 76: Also want .graphql/ .css/ .scss?

lib/config/rules/jest.js
```diff
@@ -17,4 +17,16 @@ module.exports = {
   'jest/prefer-to-be-undefined': 'error',
   // Ensure `expect()` is called with a single argument and there is an actual expectation made.
   'jest/valid-expect': 'error',
+  // Suggest using expect.assertions() OR expect.hasAssertions()
+  'jest/valid-expect-in-promise': 'off',
```
lemonmade: I think you are missing a rule (the above us not the correct description for the rule)

lemonmade: Also, this rule should definitely be on, it’s a very common problem

lib/config/rules/promise.js
```diff
@@ -22,6 +22,10 @@ module.exports = {
   'promise/no-callback-in-promise': 'off',
   // Avoid creating new promises outside of utility libs (use pify instead)
   'promise/avoid-new': 'off',
+  // Ensures the proper number of arguments are passed to Promise functions
+  'promise/valid-params': 'error',
+  // Avoid calling new on a Promise static method
+  'promise/no-new-statics': 'off',
```
lemonmade, line 28: I see no reason to leave this off

lib/config/rules/react.js
```diff
@@ -49,6 +51,8 @@ module.exports = {
   'react/no-render-return-value': 'error',
   // Prevent usage of setState
   'react/no-set-state': 'off',
+  // Prevent this from being used in stateless functional components
+  'react/no-this-in-sfc': 'off',
```
lemonmade: Same here — this is a great rule

lib/config/rules/react.js
```diff
@@ -92,6 +96,8 @@ module.exports = {
 
   // Enforce boolean attributes notation in JSX
   'react/jsx-boolean-value': 'error',
+  // Enforce or disallow spaces inside of curly braces in JSX attributes and expressions
+  'react/jsx-child-element-spacing': 'off',
```
lemonmade: I think we’d want this rule on as well

lemonmade (commented):
This is great! Do we have to do something to have our own rules support he new doc links thing?

lemonmade:
All the Jest rules are good but I think we need to disable the lowercase one for now until we can turn it off for describe, as we recommend using the full component/ class name there where appropriate.

lib/config/rules/promise.js
kaelig: It’s already handled by another rule (`no-new`)

lib/config/rules/react.js
kaelig: If Prettier already takes care of this for devs, I’m not sure this needs to be turned on

lemonmade: I don't know if Prettier would take care of this? It's more of a clarification mechanism for developers, and in any case, Prettier is not necessarily being used when using this particular set of rules (if it does conflict with Prettier, we can turn it off just for the prettier config)

.circleci/config.yml
kaelig: Yeah I got a bit carried away 😅

lib/config/rules/import.js
kaelig: Thanks, addressed in d2f1785

lib/config/rules/jest.js
kaelig: Good catch, it was missing https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/prefer-expect-assertions.md

kaelig: > Also, this rule should definitely be on, it’s a very common problem

Turned on in 23be4b8

lib/config/rules/react.js
kaelig: Turned on in f18da27

lib/config/rules/react.js
kaelig: Oh you're right, I was thinking of another rule that had to do with styles and has almost the same name. Turned on in 9978763

kaelig:
Thank you for the feedback!

Recap of the changes:
-  Allow graphql, css, sass, scss, less, styl file extensions in imports
- Turn rule `jest/valid-expect-in-promise` on
- Turn rule `react/no-this-in-sfc` on
- Turn rule `react/jsx-child-element-spacing` on
- Turn rule `jest/lowercase-name` off (for now)

---

How do we feel about this rule https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/consistent-test-it.md?

---

> This is great! Do we have to do something to have our own rules support he new doc links thing?

Yes we do. I'm opening a new branch for that.

ismail-syed:
When I wrote `jest/lowercase-name `, I originally had it to only flag `it` descriptions. The maintainers suggested enforcing `describe` and `test`. https://github.com/jest-community/eslint-plugin-jest/pull/56#discussion_r161391835.

Glad to see there's an open [PR](https://github.com/jest-community/eslint-plugin-jest/pull/92) for adding ignore options

kaelig:
Just published: `20.0.0-beta.4` 🚢 

@lemonmade `react/jsx-child-element-spacing` conflicts with prettier:

screenshot 2018-03-14 14 49 27
> Replace `{'·'}` with `·` (prettier/prettier)

---

screenshot 2018-03-14 14 49 46
> Ambiguous spacing after previous element span

---

I agree this is a useful rule, any suggestions on how to deal with it?

Edit: found a workaround:

```jsx
                <span className={styles.HeaderTextShopify}>Shopify</span>{' '}
                Polaris
```

lemonmade:
I still might be misunderstanding the rule, but I don't think they conflict. The biggest overlap is that Prettier might sometimes change code that this rule would flag to not require the explicit space at all (example from the rule):

```ts
<div>
  Here is a
  <a>link</a>
</div>

// becomes with Prettier:
<div>Here is a  <a>link</a></div>

// which is acceptable for this rule, which deals only with newline-separate siblings I think
```

I confirmed that Prettier does not try to correct the case where you make explicitly no space between newline-separate children. I think we are fine to leave this on, and it just kind of works in tandem with the Prettier formatting.

kaelig:
Thanks, I guess I was a bit confused because you can't add a trailing space into an element (`<Foo>foo </Foo>` makes ESLint and Prettier conflict) but it's so rarely a thing that I don't think it's an issue.

Let's ship it!

PR #86, opened 2018-03-14
Rewrite rules in imperative tone

kaelig:


lib/config/rules/typescript.js
```diff
@@ -1,50 +1,50 @@
...
 
 module.exports = {
-  // enforces one space after the colon and zero spaces before the colon of a type annotation.
+  // Enforce one space after the colon and zero spaces before the colon of a type annotation.
   'typescript/type-annotation-spacing': ['error'],
 
   // Require explicit return types on functions and class methods
   'typescript/explicit-function-return-type': 'off',
 
-  // enforces accessibility modifiers on class properties and methods. (member-access from TSLint)
```
lemonmade, line 10: I think these were usually just copied from plugin docs, I like the consistency though.

lemonmade (approved):

PR #89, opened 2018-03-21
Turn off unnecessary/ bad rules

lemonmade:
As we are slowly [turning on all ESLint rules in web](https://github.com/Shopify/web/issues/2073), a few have stuck out as being bad/ unnecessary. This PR turns them off, and I have commented on why I feel they should be turned off inline.

### DIFF
--- a/lib/config/node.js
+++ b/lib/config/node.js
@@ -1,3 +1,5 @@
+const merge = require('merge');
+
 module.exports = {
   env: {
     node: true,
@@ -5,5 +7,7 @@ module.exports = {
 
   plugins: ['node'],
 
-  rules: require('./rules/node'),
+  rules: merge(require('./rules/node'), {
+    'no-process-env': 'off',
+  }),
 };
--- a/lib/config/react.js
+++ b/lib/config/react.js
@@ -21,7 +21,6 @@ module.exports = {
 
   rules: merge(require('./rules/react'), require('./rules/jsx-a11y'), {
     'shopify/react-initialize-state': 'error',
-    'shopify/react-type-state': 'error',
     'shopify/jsx-no-complex-expressions': 'error',
   }),
 };
--- a/lib/config/rules/promise.js
+++ b/lib/config/rules/promise.js
@@ -9,7 +9,7 @@ module.exports = {
   // Avoid wrapping values in Promise.resolve or Promise.reject when not needed
   'promise/no-return-wrap': 'error',
   // Ensure that inside a then() you make sure to return a new promise or value.
-  'promise/always-return': 'error',
+  'promise/always-return': 'off',
   // Enforce standard parameter names for Promise constructors.
   'promise/param-names': 'error',
   // Ensure that Promise is included fresh in each file instead of relying on the existence of a native promise implementation.
--- a/lib/config/rules/react.js
+++ b/lib/config/rules/react.js
@@ -34,9 +34,9 @@ module.exports = {
   // Prevent usage of deprecated methods
   'react/no-deprecated': 'error',
   // Prevent usage of setState in componentDidMount
-  'react/no-did-mount-set-state': 'error',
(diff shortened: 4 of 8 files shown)


lib/config/node.js
```diff
@@ -1,9 +1,13 @@
...
 module.exports = {
   env: {
     node: true,
   },
 
   plugins: ['node'],
 
-  rules: require('./rules/node'),
+  rules: merge(require('./rules/node'), {
+    'no-process-env': 'off',
```
lemonmade, line 11: Servers often need to respond to the environment.

lib/config/react.js
```diff
@@ -21,7 +21,6 @@ module.exports = {
 
   rules: merge(require('./rules/react'), require('./rules/jsx-a11y'), {
     'shopify/react-initialize-state': 'error',
-    'shopify/react-type-state': 'error',
```
lemonmade, line 24: We no longer need this rule, as TS has updated to make it so the that `state` is correctly checked against its type argument. I guess I should just delete the rule altogether?

lib/config/rules/promise.js
```diff
@@ -9,7 +9,7 @@ module.exports = {
   // Avoid wrapping values in Promise.resolve or Promise.reject when not needed
   'promise/no-return-wrap': 'error',
   // Ensure that inside a then() you make sure to return a new promise or value.
-  'promise/always-return': 'error',
+  'promise/always-return': 'off',
```
lemonmade, line 12: IMO it is not a bad practice to simply do a side effect-ey thing after a promise resolves. This rule also creates inconsistencies in what is allowed for promises versus what is allowed for async await (we do not require those functions to return a value, even though they are just promises under the hood)

lib/config/rules/react.js
```diff
@@ -34,9 +34,9 @@ module.exports = {
   // Prevent usage of deprecated methods
   'react/no-deprecated': 'error',
   // Prevent usage of setState in componentDidMount
-  'react/no-did-mount-set-state': 'error',
+  'react/no-did-mount-set-state': 'off',
   // Prevent usage of setState in componentDidUpdate
-  'react/no-did-update-set-state': 'error',
+  'react/no-did-update-set-state': 'off',
```
lemonmade: In both cases there is often no alternative for updating in response to prop changes/ doing something only on the client. This will be less of an issue with the new props from state stuff, but I don't feel these are particularly bad practices.

lib/config/rules/variables.js
```diff
@@ -20,7 +20,7 @@ module.exports = {
   // Disallow use of undeclared variables unless mentioned in a /*global */ block
   'no-undef': 'error',
   // Disallow use of undefined variable
-  'no-undefined': 'error',
+  'no-undefined': 'off',
```
lemonmade, line 23: I think this is a kind of antiquated rule. It is also very hard to follow in TS when we often type things as being optional for which `undefined` is the only allowable value to "unset" the field.

lib/config/webpack.js
```diff
@@ -1,5 +1,7 @@
 const merge = require('merge');
 
 module.exports = {
-  rules: merge(require('./rules/webpack')),
+  rules: merge(require('./rules/webpack'), {
+    'no-process-env': 'off',
```
lemonmade, line 5: Using this in webpack helps actually kill code that only applies in a particular environment.

lib/config/react.js
ismail-syed: 👍 for 🔥 the rule - no need to confuse our future selves of this not being used.

ismail-syed (approved):
👍 LGTM. 

Can we document these changes in the `CHANGELOG.md`, please. Not why we're changing them but the fact that we've disabled them.

kaelig (approved):
Looks good to me (please update the changelog before merging).

PR #90, opened 2018-04-09
Adds option to check DOM/ Web Component props for hardcoded content

lemonmade:
This PR adds a new option, `dom`, to the `jsx-no-hardcoded-content` rule. It allows the specification of custom validation logic for DOM elements and Web Components, and provides a few sensible defaults.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+### Added
+* `shopify/jsx-no-hardcoded-content` now accepts a `dom` option that allows specifying attributes on DOM elements and Web Components to be checked for hardcoded content.
+
 ### Removed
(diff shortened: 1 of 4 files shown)


kaelig (approved):
Loving this ❤️

TzviPM (approved):
I like this very much! I'm curious, however, about how we'll handle indirect hardcoded content? For example, `MyComponent` takes a prop` foo` and renders an `a` tag with `title={this.props.foo}`, and then someone passes in `foo="some hardcoded content"` into `MyComponent`?

lemonmade:
@tzvipm you would have to customize the hardcoded config for `MyComponent` in that case. In the end there is only so far we can get with this, it is fairly trivial to just throw some hardcoded content in a variable and use that if you really want to get around this rule.

Issue #77, opened 2018-01-26
Add more React and Enzyme rules


lemonmade:
* No useless wrappers (`<>{stuff}{andThings}</>` > `<div>{stuff}{andThings}</div>`)
* React component should make every member private except React API stuff
* Enzyme no bad features (`setState`, `instance`, etc)
* Don't reach across component boundaries (good: `./components`, `../components`, `components`, `../OtherComponent`. bad: `../OtherComponent/any-path`, `./components/SomeComponent`)
* Avoid `renderFoo` methods being explicitly called in `render()` (use components instead)
* Avoid using most raw DOM manipulation
* Avoid the word `correct` in test names

alexandcote:
Most of the issues listed above are now solve. Feel free to reopen an issue if I missed something.

PR #93, opened 2018-05-20
Add `no-vague-titles` rule

cartogram:
Addresses one of the points here https://github.com/Shopify/eslint-plugin-shopify/issues/77 by testing the `it`, `describe` and `test` statements against a simple regex that looks for the word "correct". Will fill in the docs once the general approach is approved.

### DIFF
--- a/docs/rules/jest/no-vague-titles.md
+++ b/docs/rules/jest/no-vague-titles.md
@@ -1,0 +1,53 @@
+# Prevent the usage of vague words in test statements. (no-vague-titles)
+This rule encourages more explicit test descriptions by preventing the use of the vague words, such as "correct" and "appropriate".
+
+## Rule Details
+
+Using vague words in test descriptions often fail to communicate the meaningful details of what the underlying code is meant to do. Enabling this rule will encourage developers to write more specific and readable test descriptions.
+
+Examples of **incorrect** code for this rule:
+
+```js
+it('is called with the correct parameters')
+test('renders the appropriate markup')
+describe('receives the correct props')
+
+```
+
+Examples of **correct** code for this rule:
+
+```js
+it(`is called with the user's id and password`)
+test('renders the user avatar and email')
+describe('receives the date and publishState props from the router params')
+```
+
+### `ignore`
+
+```json
+{
+  "jest/no-vague-titles": [
+    "error",
+    {
+      "ignore": ["describe", "test"]
+    }
+  ]
+}
+```
+
+This array option whitelists function names so that this rule does not report their usage as being incorrect. There are eight possible values.
+
+* `"describe"`
+* `"test"`
+* `"it"`
+* `"xdescribe"`
+* `"xtest"`
+* `"xit"`
+* `"fdescribe"`
+* `"fit"`
+
+By default, none of these options are enabled (the equivalent of `{ "ignore": [] }`).
+
+## When Not To Use It
+
+If you do not wish to prevent the use of the vague words in test descriptions, you can safely disable this rule.
--- a/lib/rules/jest/no-vague-titles.js
+++ b/lib/rules/jest/no-vague-titles.js
@@ -1,0 +1,96 @@
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Prevent the usage of vague words in test statements.',
+      category: 'Best Practices',
+      recommended: false,
+      uri:
+        'https://github.com/Shopify/eslint-plugin-shopify/blob/master/docs/rules/no-vague-titles.md',
+    },
+  },
+  create(context) {
+    const ignored = (context.options[0] && context.options[0].ignore) || [];
+
+    function isIgnoredFunctionName({callee}) {
+      return ignored.some(
+        (method) =>
+          callee.name ? method === callee.name : method === callee.object.name,
+      );
+    }
+
+    function validate(node) {
+      if (
+        notTestFunction(node) ||
+        isIgnoredFunctionName(node) ||
+        hasEmptyDescription(node)
+      ) {
+        return;
+      }
+
+      const description = getDescription(node);
+
+      if (containsVagueWord(description)) {
+        context.report({
+          message: `{{ method }} description should not contain vague words. Be sure the description meaningfully illustrates the purpose of this test.`,
+          data: {
+            method: node.callee.name
+              ? node.callee.name
+              : node.callee.object.name,
+          },
+          node,
+        });
+      }
+    }
+
+    return {
+      CallExpression(node) {
+        validate(node);
+      },
+    };
+  },
+};
+
+function notTestFunction({callee}) {
+  return callee.name
+    ? !matchTestFunctionName(callee.name)
+    : !matchTestFunctionName(callee.object.name);
+}
+
+function matchTestFunctionName(functionName) {
+  return (
+    functionName === 'it' ||
+    functionName === 'xit' ||
+    functionName === 'fit' ||
+    functionName === 'test' ||
+    functionName === 'xtest' ||
+    functionName === 'describe' ||
+    functionName === 'fdescribe' ||
+    functionName === 'xdescribe'
+  );
+}
+
+function hasEmptyDescription({arguments: args}) {
+  return (
+    !args ||
+    !args[0] ||
+    (args[0].type !== 'Literal' && args[0].type !== 'TemplateLiteral')
+  );
+}
+
+function getDescription({arguments: args}) {
+  const firstArgument = args[0];
+
+  if (firstArgument.type === 'TemplateLiteral') {
+    return firstArgument.quasis
+      .map((templateLiteral) => {
+        return templateLiteral.value.raw;
+      })
+      .join('');
+  }
+
+  return firstArgument && firstArgument.value;
+}
+
+function containsVagueWord(description) {
+  return description.match(/correct/i) || description.match(/appropriate/i);
+}
--- a/tests/lib/rules/jest/no-vague-titles.js
+++ b/tests/lib/rules/jest/no-vague-titles.js
@@ -1,0 +1,602 @@
+const {RuleTester} = require('eslint');
+const rule = require('../../../../lib/rules/jest/no-vague-titles');
+
+const ruleTester = new RuleTester();
+
+require('babel-eslint');
+
+const parser = 'babel-eslint';
+
+function errorWithMethod(method) {
+  return [
+    {
+      type: 'CallExpression',
+      message: `${method} description should not contain vague words. Be sure the description meaningfully illustrates the purpose of this test.`,
+    },
+  ];
+}
+ruleTester.run('no-vague-titles', rule, {
+  valid: [
+    {
+      code: `it()`,
+      parser,
+    },
+    {
+      code: `it('')`,
+      parser,
+    },
+    {
+      code: `it('foo bar baz')`,
+      parser,
+    },
+    {
+      code: `xit('foo bar baz')`,
+      parser,
+    },
+    {
+      code: `it.only('foo bar baz')`,
+      parser,
+    },
+    {
+      code: `describe('foo bar baz')`,
+      parser,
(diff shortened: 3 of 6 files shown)


cartogram:
Should we nest this rule inside of a "jest" folder (like with "webpack")?

GoodForOneFare:
So much potential 😍 Some suggestions:
- Rename to be more general (e.g., `no-weasel-words`, `no-ambiguous-titles`, `no-vague-titles`)
- Forbid `appropriate`
- Forbid double-negatives
- Submit this to [`eslint-plugin-jest`](https://github.com/jest-community/eslint-plugin-jest)

The double-negatives thing is maybe its own rule, but it could reuse code from this one.

Mocha has a similar rule: https://github.com/lo1tuma/eslint-plugin-mocha/blob/master/docs/rules/valid-test-description.md#rule-details.

lib/rules/no-tests-contain-correct.js
```diff
@@ -0,0 +1,81 @@
...
+      uri:
+        'https://github.com/Shopify/eslint-plugin-shopify/blob/master/docs/rules/no-tests-contain-correct.md',
+    },
+  },
+  create(context) {
+    const ignore = (context.options[0] && context.options[0].ignore) || [];
+    const ignoredFunctionNames = ignore.reduce((accumulator, value) => {
+      accumulator[value] = true;
+      return accumulator;
+    }, Object.create(null));
```
lemonmade: Probably better as a map, a little easier to construct too

lib/rules/no-tests-contain-correct.js
```diff
@@ -0,0 +1,81 @@
...
+  return (
+    node.type === 'CallExpression' &&
+    node.callee &&
+    (node.callee.name === 'it' ||
+      node.callee.name === 'test' ||
+      node.callee.name === 'describe')
+  );
+}
+
+function isItDescription(node) {
```
lemonmade: This function name is confusing to me

lib/rules/no-tests-contain-correct.js
```diff
@@ -0,0 +1,81 @@
...
+    return firstArgument.value;
+  }
+
+  return firstArgument.quasis[0].value.raw;
+}
+
+function descriptionContainsCorrect(node) {
+  if (isItTestOrDescribeFunction(node) && isItDescription(node)) {
+    const description = testDescription(node);
+    if (!description[0]) {
```
lemonmade: Checking if empty string?

lib/rules/no-tests-contain-correct.js
```diff
@@ -0,0 +1,81 @@
...
+      return false;
+    }
+
+    if (description.match(/correct/gi)) {
+      return node.callee.name;
```
lemonmade: Why return the name?

lib/rules/no-tests-contain-correct.js
```diff
@@ -0,0 +1,81 @@
...
+
+    function isIgnoredFunctionName(node) {
+      return ignoredFunctionNames[node.callee.name];
+    }
+
+    return {
+      CallExpression(node) {
+        const erroneousMethod = descriptionContainsCorrect(node);
+
+        if (erroneousMethod && !isIgnoredFunctionName(node)) {
```
lemonmade: Seems like a weird flow to me — I'd expect it to check if this is one of the functions we expect, removing ignored ones, then check if the description contains invalid words, then report.

lib/rules/no-tests-contain-correct.js
```diff
@@ -0,0 +1,81 @@
...
+            data: {method: erroneousMethod},
+            node,
+          });
+        }
+      },
+    };
+  },
+};
+
+function isItTestOrDescribeFunction(node) {
```
lemonmade: Probably good to catch the skipped versions (`fit`/ `it.only` and the same for the other kinds)

lemonmade (changes requested):
Like Gord's suggestion of expanding this to be more of a "no-vague-titles" rule. Also all for nesting under `jest/`

cartogram:
@lemonmade @GoodForOneFare thanks for the feedback! I have addressed everything and think this is ready for another look.

docs/rules/jest/no-vague-titles.md
```diff
@@ -0,0 +1,53 @@
+# Prevent the usage of vague words in test statements.(no-vague-titles)
+This rule prevents the use of the words "correct" and "appropriate" in test descriptions.
```
lemonmade: Make this a little more general

docs/rules/jest/no-vague-titles.md
```diff
@@ -0,0 +1,53 @@
+# Prevent the usage of vague words in test statements.(no-vague-titles)
```
lemonmade: Space after period

docs/rules/jest/no-vague-titles.md
```diff
@@ -0,0 +1,53 @@
...
+This rule prevents the use of the words "correct" and "appropriate" in test descriptions.
+
+## Rule Details
+
+Using vague words to convey the valid test path within `it`, `describe` or `test` descriptions often fails to communicate the meaningful details of what the underlying code is meant to do. Enabling this rule will encourage developers to write more specific and readable test descriptions.
```
lemonmade: Not sure what `to convey the valid test path within` means?

docs/rules/jest/no-vague-titles.md
```diff
@@ -0,0 +1,53 @@
...
+
+Examples of **incorrect** code for this rule:
+
+```js
+
```
lemonmade: Why do you leave empty lines in these?

lib/rules/jest/no-vague-titles.js
```diff
@@ -0,0 +1,92 @@
...
+      if (node.type === 'MemberExpression') {
+        return ignored.some((method) => method === node.object.name);
+      }
+      return ignored.some((method) => method === node.callee.name);
+    }
+
+    function validate(node) {
+      if (
+        !isTestFunction(node) ||
+        isIgnoredFunctionName(node) ||
```
lemonmade: I would probably flip this condition so they are all either positive or negative

lib/rules/jest/no-vague-titles.js
```diff
@@ -0,0 +1,92 @@
...
+        !containsDescription(node)
+      ) {
+        return;
+      }
+
+      const description = getDescription(node);
+
+      if (containsVagueWord(description)) {
+        context.report({
+          message: `{{ method }} statement should not contain vague words`,
```
lemonmade: description instead of statement

lib/rules/jest/no-vague-titles.js
```diff
@@ -0,0 +1,92 @@
...
+          data: {method: node.callee ? node.callee.name : node.object.name},
+          node,
+        });
+      }
+    }
+
+    return {
+      MemberExpression(node) {
```
lemonmade: Why do you need this one? I think we only care about call expressions.

lib/rules/jest/no-vague-titles.js
```diff
@@ -0,0 +1,92 @@
...
+
+  return node.callee && matchTestFunctionName(node.callee.name);
+}
+
+function matchTestFunctionName(functionName) {
+  return (
+    functionName === 'it' ||
+    functionName === 'fit' ||
+    functionName === 'test' ||
+    functionName === 'describe'
```
lemonmade: I think `fdescribe` also works, `test` might also have that form

lib/rules/jest/no-vague-titles.js
```diff
@@ -0,0 +1,92 @@
...
+function getDescription(node) {
+  const args = node.arguments || node.parent.arguments || [];
+  const firstArgument = args[0];
+  const type = firstArgument.type;
+
+  if (type === 'Literal') {
+    return firstArgument.value;
+  }
+
+  return firstArgument.quasis[0].value.raw;
```
lemonmade: This is a little odd, might want to merge together all the quasis that you can to come up with a complete description

lib/rules/jest/no-vague-titles.js
```diff
@@ -0,0 +1,92 @@
...
+  );
+}
+
+function getDescription(node) {
+  const args = node.arguments || node.parent.arguments || [];
+  const firstArgument = args[0];
+  const type = firstArgument.type;
+
+  if (type === 'Literal') {
+    return firstArgument.value;
```
lemonmade: What if it's not a string literal?

lib/rules/jest/no-vague-titles.js
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  return firstArgument.quasis[0].value.raw;
+}
+
+function containsVagueWord(description) {
+  return description.match(/correct/giu) || description.match(/appropriate/giu);
```
lemonmade: No need for `g`, we also don't usually bother with `u`

tests/lib/rules/jest/no-vague-titles.js
```diff
@@ -0,0 +1,390 @@
...
+    {
+      code: `it('foo bar baz')`,
+      parser,
+    },
+    {
+      code: `describe('foo bar baz')`,
+      parser,
+    },
+    {
+      code: `test('foo bar baz')`,
```
lemonmade: You should also test non-test functions called with an argument that contains the bad words

tests/lib/rules/jest/no-vague-titles.js
```diff
@@ -0,0 +1,390 @@
...
+      parser,
+      errors: errorWithMethod('it', 'MemberExpression'),
+    },
+    {
+      code: "describe('correct')",
+      parser,
+      errors: errorWithMethod('describe'),
+    },
+    {
+      code: "describe('correct')",
```
lemonmade, line 172: I would ignore for now, but you might be interested in tackling the new `describe/it/test.each` Jest recently introduced: https://facebook.github.io/jest/docs/en/api.html#describeeachtable-name-fn

lib/rules/jest/no-vague-titles.js
lemonmade: I think it would also be nice to briefly mention what they should do instead; write a description that more meaningfully illustrates the purpose of the test/ suite.

lemonmade (changes requested):

lib/rules/jest/no-vague-titles.js
cartogram: @lemonmade `it.only` vs `it`. See below:

screen shot 2018-06-02 at 5 39 59 pm

lib/rules/jest/no-vague-titles.js
cartogram: `ftest` doesn't exist, but there is `xtest` as well as `xdescribe`, `xit`. Will add all of these.

cartogram:
Thanks again @lemonmade! Last round should all be addressed, let me know if there is anything else or good to merge.

lib/rules/jest/no-vague-titles.js
lemonmade: But `validate` always works for call expressions, regardless of whether the callee is just an identifier or a member expression. You end up basically validating every member expression twice (once at the call expression, once at the member expression)

lib/rules/jest/no-vague-titles.js
```diff
@@ -0,0 +1,99 @@
...
+  const args = node.arguments || node.parent.arguments || [];
+  return (
+    !args ||
+    !args[0] ||
+    (args[0].type !== 'Literal' && args[0].type !== 'TemplateLiteral')
+  );
+}
+
+function getDescription(node) {
+  const args = node.arguments || node.parent.arguments || [];
```
lemonmade: This is only needed because you are unnecessarily validating the member expression

lemonmade (approved):
Still think you can skip the member expression check.

PR #94, opened 2018-05-20
Add `jsx-prefer-fragment-wrappers` rule

cartogram:
Addresses one of the points here #77 by testing JSX code for useless wrapping elements. Will fill in the docs once the general approach is approved.

### DIFF
--- a/docs/rules/jsx-prefer-fragment-wrappers.md
+++ b/docs/rules/jsx-prefer-fragment-wrappers.md
@@ -1,0 +1,39 @@
+# Disallow useless wrapping elements in favour of fragment shorthand in JSX. (jsx-prefer-fragment-wrappers)
+
+It is common in React for a component to return multiple elements. React Fragments provide a way to wrap a list of child nodes without adding an extra div to the DOM.
+
+## Rule Details
+
+This rule prevents the use of additional wrapping elements and enforces the use of React Fragments instead.
+
+Examples of **incorrect** code for this rule:
+
+```js
+render() {
+  return (
+    <div>
+      <ChildA />
+      <ChildB />
+      <ChildC />
+    </div>
+  );
+}
+```
+
+Examples of **correct** code for this rule:
+
+```js
+render() {
+  return (
+    <>
+      <ChildA />
+      <ChildB />
+      <ChildC />
+    </>
+  );
+}
+```
+
+## When Not To Use It
+
+If you do not wish to enforce the use of React Fragments, you can safely disable this rule.
--- a/lib/rules/jsx-prefer-fragment-wrappers.js
+++ b/lib/rules/jsx-prefer-fragment-wrappers.js
@@ -1,0 +1,57 @@
+const pascalCase = require('pascal-case');
+
+module.exports = {
+  meta: {
+    docs: {
+      description:
+        'Disallow useless wrapping elements in favour of fragment shorthand in JSX',
+      category: 'Best Practices',
+      recommended: false,
+      uri:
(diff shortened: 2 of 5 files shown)


lib/rules/jsx-prefer-fragment-wrappers.js
```diff
@@ -0,0 +1,47 @@
...
+    };
+  },
+};
+
+function hasOneOrNoChildren({children}) {
+  const childNodes = children.filter((child) => child.type !== 'Literal');
+  return childNodes.length < 2;
+}
+
+function isComponent({openingElement: {name}}) {
```
lemonmade: This will not correctly catch things like `<Foo.Bar>`

lib/rules/jsx-prefer-fragment-wrappers.js
```diff
@@ -0,0 +1,47 @@
...
+  return name.type === 'JSXIdentifier' && /^[A-Z]/.test(name.name);
+}
+
+function hasAttributes({openingElement: {attributes}}) {
```
lemonmade: Not my favourite to return a false-ey value here instead of a true boolean

tests/lib/rules/jsx-prefer-fragment-wrappers.js
```diff
@@ -0,0 +1,41 @@
...
+      type: 'JSXElement',
+      message: `replace wrapping ${tagName} with fragment shorthand`,
+    },
+  ];
+}
+
+ruleTester.run('jsx-prefer-fragment-wrappers', rule, {
+  valid: [
+    {code: '<div className={className}>{foo}{bar}<Baz /></div>', parserOptions},
+    {code: '<div><Bar /></div>', parserOptions},
```
lemonmade: This also seems useless?

tests/lib/rules/jsx-prefer-fragment-wrappers.js
```diff
@@ -0,0 +1,41 @@
...
+    {code: '<Foo><Bar /><Baz /></Foo>', parserOptions},
+    {code: '<Foo>{someFunction()}</Foo>', parserOptions},
+    {code: '<Foo>Some content</Foo>', parserOptions},
```
lemonmade: Test with a real fragment?

lemonmade (changes requested):
A few little comment but overall looks great

tests/lib/rules/jsx-prefer-fragment-wrappers.js
cartogram: @lemonmade Yeah I suppose it is, but I would argue that this is still valid for this rule. Right now, I'm checking for multiple children because we are trying to prevent the case where a developer will add a wrapping `<div>` instead of a Fragment. If we are just checking for useless markup, that might be a different rule or at least a different error message.

cartogram:
@lemonmade this is ready for another look. 🙏

docs/rules/jsx-prefer-fragment-wrappers.md
```diff
@@ -0,0 +1,39 @@
...
+
+```js
+render() {
+  return (
+    <React.Fragment>
+      <ChildA />
+      <ChildB />
+      <ChildC />
+    </React.Fragment>
+  );
```
lemonmade: Can we show the shorthand instead?

tests/lib/rules/jsx-prefer-fragment-wrappers.js
```diff
@@ -0,0 +1,49 @@
...
+    {
+      code: '<React.Fragment><Foo /><Bar /><Baz /></React.Fragment>',
```
lemonmade: Test with fragment syntax too

lemonmade (approved):

PR #95, opened 2018-05-20
Add `react-prefer-private-members` rule

cartogram:
Addresses one of the points here https://github.com/Shopify/eslint-plugin-shopify/issues/77 by looking for React class components and making sure they do not have private members. React API stuff will pass through. Will fill in the docs once the general approach is approved.

### DIFF
--- a/docs/rules/react-prefer-private-members.md
+++ b/docs/rules/react-prefer-private-members.md
@@ -1,11 +1,10 @@
 # Disallow public members within React component classes (react-prefer-private-members)
 
-Please describe the origin of the rule here.
-
+This rule enforces all non-React-specific members be marked private in React class components.
 
 ## Rule Details
 
-This rule aims to...
+When a member of class is marked private, it cannot be accessed from outside of its containing class. This restriction is prefered when writing React components, where encapsulation is almost always desireable.
 
 Examples of **incorrect** code for this rule:
 
@@ -23,14 +22,6 @@ Examples of **correct** code for this rule:
 
 ```
 
-### Options
-
-If there are any options, describe them here. Otherwise, delete this section.
-
 ## When Not To Use It
 
 Give a short description of when it would be appropriate to turn off this rule.
-
-## Further Reading
-
-If there are other links that describe the issue this rule addresses, please include them here in a bulleted list.
--- a/docs/rules/react-prefer-private-members.md
+++ b/docs/rules/react-prefer-private-members.md
@@ -1,0 +1,36 @@
+# Disallow public members within React component classes (react-prefer-private-members)
+
+Please describe the origin of the rule here.
+
+
+## Rule Details
+
+This rule aims to...
+
+Examples of **incorrect** code for this rule:
+
+```js
+
+// fill me in
+
+```
+
+Examples of **correct** code for this rule:
+
+```js
+
+// fill me in
+
+```
+
+### Options
+
+If there are any options, describe them here. Otherwise, delete this section.
+
+## When Not To Use It
+
+Give a short description of when it would be appropriate to turn off this rule.
+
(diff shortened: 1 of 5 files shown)


lib/rules/react-prefer-private-members.js
```diff
@@ -0,0 +1,85 @@
+const Components = require('eslint-plugin-react/lib/util/Components');
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Disallow public members within React component classes',
+      category: 'Possible Errors',
```
cartogram: Not really sure what to put in here. Just did my best to follow other rules in this repo.

lib/rules/react-prefer-private-members.js
```diff
@@ -0,0 +1,85 @@
...
+      recommended: true,
```
cartogram: Same with this.

lib/rules/react-prefer-private-members.js
lemonmade: It's probably best practices

lib/rules/react-prefer-private-members.js
```diff
@@ -0,0 +1,85 @@
...
+  }),
+};
+
+function makeReport(node, {componentName}) {
+  const {
+    key: {name},
+  } = node;
+  return {
+    node,
+    message: `${name} should be a private member of ${componentName}.`,
```
lemonmade: You don't need to pass the variables?

lib/rules/react-prefer-private-members.js
```diff
@@ -0,0 +1,85 @@
...
+    'UNSAFE_componentWillUpdate',
+    'getSnapshotBeforeUpdate',
+    'componentDidUpdate',
+    'componentDidCatch',
+    'componentWillUnmount',
+    'render',
+  ].some((method) => method === name);
+}
+
+function isReactStaticProperty({key: {name}}) {
```
lemonmade, line 82: I think you probably want to allow all static properties, or at least those that are pascal-case, to allow for our (pretty common) pattern of exposing subcomponent as public static members.

tests/lib/rules/react-prefer-private-members.js
```diff
@@ -0,0 +1,87 @@
...
+        getSnapshotBeforeUpdate() {}
+        componentDidUpdate() {}
+        componentDidCatch() {}
+        componentWillUnmount() {}
+        render() {}
+      }`,
+      parser: babelParser,
+    },
+  ],
+  invalid: [
```
lemonmade, line 72: Need more tests than this (missing for private members, static members, etc

lemonmade (approved):
Approach looks great!

cartogram:
@lemonmade this is ready for another look. 🙏

docs/rules/react-prefer-private-members.md
```diff
@@ -0,0 +1,66 @@
...
+
+The following patterns are not warnings:
+```ts
+class MyComponent extends React.Component<Props, State> {
+  private publicMethod() {}
+  private alsoPublic: string;
+}
+```
+
+The Life Cycle methods and static properties that are part of the React API are not required to be private for this rule.
```
lemonmade: `lifecycle`

docs/rules/react-prefer-private-members.md
```diff
@@ -0,0 +1,66 @@
...
+
+```ts
+class MyComponent extends React.Component<Props, State> {
+  static MySubComponent = MySubComponent;
+  render() {}
+}
+```
+
+
+
```
lemonmade: So many empty lines :D

lemonmade (approved):

PR #98, opened 2018-06-10
Add `strict-component-boundaries` rule

cartogram:
Another one off this list https://github.com/Shopify/eslint-plugin-shopify/issues/77

### DIFF
--- a/docs/rules/strict-component-boundaries.md
+++ b/docs/rules/strict-component-boundaries.md
@@ -1,0 +1,25 @@
+# Prevent module imports between components. (strict-component-boundaries)
+
+Components should have everything they need co-located in a single directory. This makes it easy to move components to a more appropriate spot, and easy for future developers to find, understand, and update everything about a component in one place. To help maintain encapsulation, components should not import any part of another component, and should be restrictive about the parts of their own API that are exported.
+
+## Rule Details
+
+This rule prevents developers from importing across component directories.
+
+Examples of **incorrect** code for this rule:
+
+```js
+import {someThing} from './components/SomeComponent';
+import someThing from './SomeComponent/anything';
+```
+
+Examples of **correct** code for this rule:
+
+```js
+import {someThing} from './components';
+import someThing from 'utilities/anything';
+```
+
+## When Not To Use It
+
+If you do not wish to enforce strict component boundaries, you can safely disable this rule.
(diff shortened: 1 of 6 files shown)


docs/rules/strict-component-boundaries.md
```diff
@@ -0,0 +1,25 @@
+# Prevent module imports between components. (strict-component-boundaries)
+
+Components should have everything they need co-located in a single directory. This makes it easy to move components to a more appropriate spot, and makes it easy for future developers to find, understand, and update everything about a component in one place.
```
lemonmade: I think need a comment that other components should not be reaching in/ the component should export its entire API

docs/rules/strict-component-boundaries.md
```diff
@@ -0,0 +1,25 @@
...
+
+```js
+import {something} from './components/SomeComponent';
+import something from './SomeComponent/anything';
+```
+
+Examples of **correct** code for this rule:
+
+```js
+import {somthing} from './components';
```
lemonmade: Missing an e, make it pascalcase

tests/lib/rules/strict-component-boundaries.js
```diff
@@ -0,0 +1,42 @@
...
+  valid: [
+    {code: `import {something} from './components';`, parserOptions},
+    {
+      code: `import {something} from 'components';`,
+      parserOptions,
+    },
+  ],
+  invalid: [
+    {
+      code: `import something from '../OtherComponent/any-path';`,
```
lemonmade: Can you check that ../OtherComponent works?

lib/rules/strict-component-boundaries.js
```diff
@@ -0,0 +1,41 @@
...
+  create(context) {
+    return {
+      ImportDeclaration(node) {
+        const pathParts = node.source.value
+          .split('/')
+          .filter((part) => part[0] !== '.');
+
+        if (
+          (hasAnotherComponentInPath(pathParts) && pathParts.length > 1) ||
+          (hasComponentDirectoryInPath(pathParts) && pathParts.length > 2)
```
lemonmade, line 23: I think this might not catch ‘components/Foo’

lemonmade (changes requested):

lib/rules/strict-component-boundaries.js
cartogram: The first one should catch that https://github.com/Shopify/eslint-plugin-shopify/pull/98/files#diff-f91ed48b12fe78aaf97a3a26453bd862R31

cartogram:
Thanks for the feedback @lemonmade, it is ready for another look now

lemonmade (approved):

PR #100, opened 2018-06-16
Adds the latest rules to main JS file

cartogram:
I missed this step in the last 3 rules I merged in.


lemonmade (approved):
Do we need to add heir configuration to the different configs we have?

PR #106, opened 2018-07-03
Adds properties to `react-prefer-private-members`

cartogram:
Turning this rule on in `web` revealed a few missed properties. This PR adds them.

### DIFF
--- a/lib/rules/react-prefer-private-members.js
+++ b/lib/rules/react-prefer-private-members.js
@@ -56,6 +56,7 @@ function isValid(node) {
     node.accessibility === 'private' ||
     isReactLifeCycleMethod(node) ||
     isReactStaticProperty(node) ||
+    isConstructor(node) ||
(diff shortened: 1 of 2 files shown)


BPScott (approved):

BPScott:
Worth adding a changelog entry?

lib/rules/react-prefer-private-members.js
```diff
@@ -85,9 +89,14 @@ function isReactStaticProperty({key: {name}}) {
     'contextTypes',
     'childContextTypes',
     'defaultProps',
+    'displayName',
   ].some((method) => method === name);
 }
 
+function isOtherAcceptedProperty({key: {name}}) {
```
lemonmade: I would just check if this is a constructor, I think that also has a different node type

lemonmade (approved):

PR #107, opened 2018-07-03
Move prettier to be a peer dependency

BPScott:
Helping along with https://github.com/Shopify/sewing-kit/issues/737. We should strive to have a single prettier install underneath sewing-kit, rather than multiple dependencies potentially installing multiple prettier versions which leads to confusion around different formatting tools producing different results.

eslint-plugin-prettier expects prettier to be a peer dependency, so do the same here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,15 @@
 
 ## Unreleased
 
+eslint-plugin-shopify will no longer install prettier as a dependency. Please ensure you have added prettier to your package.json if you wish to use it.
+
 ### Added
 * `shopify/jsx-prefer-fragment-wrappers` ([#94](https://github.com/Shopify/eslint-plugin-shopify/pull/94))
 * `shopify/jest/no-vague-titles` ([#93](https://github.com/Shopify/eslint-plugin-shopify/pull/93))
 * `shopify/strict-component-boundaries` ([#98](https://github.com/Shopify/eslint-plugin-shopify/pull/98))
+
+### Changed
+* **Breaking** Moved prettier to be a peerDependency, this avoids the potential for having multiple versions of prettier in the dependency graph. If you use prettier you will need to ensure you have it installed in your project as eslint-plugin-shopify will no longer install it for you ([#107](https://github.com/Shopify/eslint-plugin-shopify/pull/107))
 
 ## [22.1.0] - 2018-06-08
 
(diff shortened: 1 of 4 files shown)


BPScott:
Currently running `yarn why prettier` in web returns:

```
info Reasons this module exists
   - "@shopify#sewing-kit" depends on it
   - Hoisted from "@shopify#sewing-kit#prettier"
   - Hoisted from "@shopify#sewing-kit#eslint-plugin-shopify#prettier"
   - Hoisted from "@shopify#sewing-kit#prettier-stylelint-formatter#prettier"
   - Hoisted from "@shopify#sewing-kit#stylelint-config-shopify#prettier-stylelint-formatter#prettier"
```

This should remove the `Hoisted from "@shopify#sewing-kit#eslint-plugin-shopify#prettier"` line. See the linked issue for PRs relating to removing the others

ismail-syed (approved):
I believe this is the right call here. 

We had discussion about this in the past. We had `prettier` set as an optional dependency because consumers may use `eslint-plugin-shopify` without the use of `prettier`. 

However, docs on `[optionalDependencies`](https://docs.npmjs.com/files/package.json#optionaldependencies) seem to imply different behaviour. 

> The difference is that build failures do not cause installation to fail.

It doesn't seem to have much impact on how dependency resolution works with regards to `devDependencies` vs `optionalDependencies`. IIRC, when I first PR'd this, I had it set as a peerDep, then changed it to an optionalDep after PR feedback.

ismail-syed:
We should update the readme to let consumers know that `prettier` should be installed separately as a peerDep.

GoodForOneFare (approved):
👍 Needs a CHANGELOG entry, ideally highlighting that this is a breaking change.

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@
 * `shopify/jest/no-vague-titles` ([#93](https://github.com/Shopify/eslint-plugin-shopify/pull/93))
 * `shopify/strict-component-boundaries` ([#98](https://github.com/Shopify/eslint-plugin-shopify/pull/98))
 
+### Changed
+* **Breaking** Moved prettier to be a peerDependency, this avoids the potential for having multiple versions of prettier in the dependency graph ([#107](https://github.com/Shopify/eslint-plugin-shopify/pull/107))
```
ismail-syed: Can we also add a note here for consumers to install prettier themselves. Frequent visitors are more likely to read the changelog over readme history.

PR #109, opened 2018-07-03
Fixes `no-vague-titles`

cartogram:
This error came up when testing in `web`. A minor fix to prevent reading `name` from a potentially undefined `callee.object`.


lemonmade (approved):
Where be the test case?

cartogram:
@lemonmade added a test case. In doing so I feel more confident it was IIFEs that was causing the error. The error output for the IIFE test case matches what I was seeing and code added in this PR fixes it.

PR #112, opened 2018-07-09
fix `jsx-prefer-fragment-wrappers`

### LINKED ISSUE #111: `jsx-prefer-fragment-wrappers` returns invalid for lots of HTML markup
Consider the following example:
```
function Table({content, otherContent}) {
 return (
   <table>
     <tbody>
       <tr>{content}</tr>
       <td>{otherContent}</td>
     </tbody>
   </table>
  )
}
```

The rule currently would consider all of those tags invalid. I think we should remove all the tag checking logic to just look for `divs` and `spans`.

@lemonmade what do you think?

cartogram:
closes https://github.com/Shopify/eslint-plugin-shopify/issues/111

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,7 +119,7 @@ This plugin provides the following custom rules, which are included as appropria
 - [jquery-dollar-sign-reference](docs/rules/jquery-dollar-sign-reference.md): Require that all jQuery objects are assigned to references prefixed with `$`.
 - [jsx-no-complex-expressions](docs/rules/jsx-no-complex-expressions.md): Disallow complex expressions embedded in in JSX.
 - [jsx-no-hardcoded-content](docs/rules/jsx-no-hardcoded-content.md): Disallow hardcoded content in JSX.
-- [jsx-prefer-fragment-wrappers](docs/rules/jsx-no-hardcoded-content.md): Disallow useless wrapping elements in favour of fragment shorthand in JSX.
+- [jsx-prefer-fragment-wrappers](docs/rules/jsx-prefer-fragment-wrappers.md): Disallow useless wrapping elements in favour of fragment shorthand in JSX.
 - [no-useless-computed-properties](docs/rules/no-useless-computed-properties.md): Prevent the usage of unnecessary computed properties.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -119,7 +119,7 @@ This plugin provides the following custom rules, which are included as appropria
 - [jquery-dollar-sign-reference](docs/rules/jquery-dollar-sign-reference.md): Require that all jQuery objects are assigned to references prefixed with `$`.
 - [jsx-no-complex-expressions](docs/rules/jsx-no-complex-expressions.md): Disallow complex expressions embedded in in JSX.
 - [jsx-no-hardcoded-content](docs/rules/jsx-no-hardcoded-content.md): Disallow hardcoded content in JSX.
-- [jsx-prefer-fragment-wrappers](docs/rules/jsx-no-hardcoded-content.md): Disallow useless wrapping elements in favour of fragment shorthand in JSX.
+- [jsx-prefer-fragment-wrappers](docs/rules/jsx-prefer-fragment-wrappers.md): Disallow useless wrapping elements in favour of fragment shorthand in JSX.
```
cartogram, line 122: unrelated typo

BPScott (approved):
LGTM.

Do we need a changelog entry?

lib/rules/jsx-prefer-fragment-wrappers.js
```diff
@@ -35,23 +33,15 @@ module.exports = {
   },
 };
 
-function isPascalCase(string) {
-  return string === pascalCase(string);
+function isAcceptedTagName({openingElement: {name}}) {
+  return ['div', 'span'].every((tagName) => tagName !== name.name);
```
lemonmade: IMO `![].some` is probably more readable here, or inverting the whole function's purpose, it is hard for me to read "accepted tag name, but we are checking that ever non-accepted tag name is not the case"

lemonmade (approved):

lib/rules/jsx-prefer-fragment-wrappers.js
cartogram: I like `![].some` and keeping this purpose consistent with the others like it in the file.

PR #114, opened 2018-07-17
Add 'all' to vague terms in no-vague-titles

ismail-syed:
I think we should also flag `all` as a vague term. Noticed this while reviewing https://github.com/Shopify/sewing-kit/pull/755.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1569,7 +1569,7 @@ eslint-plugin-react@7.7.0:
     prop-types "^15.6.0"
 
 "eslint-plugin-shopify@file:./.":
-  version "22.1.0"
+  version "23.0.0"
   dependencies:
     babel-eslint "8.2.3"
     eslint-config-prettier "2.9.0"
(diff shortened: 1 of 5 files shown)


yarn.lock
```diff
@@ -1569,7 +1569,7 @@ eslint-plugin-react@7.7.0:
     prop-types "^15.6.0"
 
 "eslint-plugin-shopify@file:./.":
-  version "22.1.0"
+  version "23.0.0"
```
ismail-syed, line 1572: `yarn install` on a fresh clone of master updates this. Our lock file is our of sync.

cartogram (commented):
Changes look great! I love this in theory, but what if we have a prop called `onAllImagesUploaded` and then in our test we have `decribe('onAllImageUploaded')?

ismail-syed:
@cartogram, good catch! I've updated the regex to only match cases where `all` is surrounded by spaces or used before a period, to help determine that he it's being used in a sentence. This prevents flagging `onAllImagesUploaded`.  I've added a test with the `onAllImagesUploaded` to verify as well. Any other cases you can think of?

cartogram (approved):
@ismail-syed brilliant! This LGTM!

PR #117, opened 2018-07-19
[strict-component-boundaries] Exclude fixtures folder

### LINKED ISSUE #116: False positive with strict-component-boundaries and fixtures
We have GraphQL fixtures in our project and we include these fixtures like that :
```import basicFixture from './fixtures/TestAdjacentResourcesQuery/basic.json';```

Because `TestAdjacentResourcesQuery` folder is PascalCase, the rule thinks it is a component but is not. We use PascalCase for that because the GraphQL file related to this fixture look like `TestAdjacentResourcesQuery.graphql`

Maybe we should exclude `fixtures` folder.

### PR DESCRIPTION
**Pull request by alexandcote**
Fix #116 

We now exclude the `fixtures` folder. Maybe we should exclude everything inside tests instead?

### CODE REVIEW

File: lib/rules/strict-component-boundaries.js [hunk 1]
```diff
@@ -19,8 +19,9 @@ module.exports = {
           .filter((part) => part[0] !== '.');
 
         if (
-          (hasAnotherComponentInPath(pathParts) && pathParts.length > 1) ||
-          (hasComponentDirectoryInPath(pathParts) && pathParts.length > 2)
+          isANotFixtureDirectory(pathParts) &&
+          ((hasAnotherComponentInPath(pathParts) && pathParts.length > 1) ||
```
cartogram: `isANotFixtureDirectory` => `isNotAFixtureDirectory`

cartogram (commented):
@alexandcote thanks for working on this! The test looks good, but what about adding a check for imports with `json` extensions instead? We are never going to want this rule to flag `json` imports because they will never be a component.

alexandcote:
@cartogram  Do we have any other file extensions that we want to exclude, `scss`, `json`, `svg`, etc.. ?

lib/rules/strict-component-boundaries.js
```diff
@@ -39,3 +40,9 @@ function hasComponentDirectoryInPath(pathParts) {
 function hasAnotherComponentInPath(pathParts) {
   return Boolean(pathParts.filter((part) => part === pascalCase(part)).length);
 }
+
+function isASupportedFileType(pathParts) {
```
cartogram: `isASupportedFileType` => `isSupportedFileType`

lib/rules/strict-component-boundaries.js
```diff
@@ -39,3 +40,9 @@ function hasComponentDirectoryInPath(pathParts) {
...
+  const unSupportedType = ['.json', '.svg', '.scss'];
```
cartogram: Unless there is a reason not to, just keep the `.json` for now.

tests/lib/rules/strict-component-boundaries.js
```diff
@@ -25,6 +25,18 @@ ruleTester.run('strict-component-boundaries', rule, {
       code: `import {someThing} from '../OtherComponent';`,
       parserOptions,
     },
+    {
+      code: `import someThing from '../NameOfTheQuery/mock-query.json';`,
+      parserOptions,
+    },
+    {
+      code: `import someThing from '../NameOfTheQuery/mock-query.scss';`,
```
cartogram: These are also bad practice so its weird to me they are `valid`. As I commented above, I think we just keep the `.json` one to allow for fixtures but I'd rather not loosen this rule more than needed.

lib/rules/strict-component-boundaries.js
```diff
@@ -19,8 +19,9 @@ module.exports = {
           .filter((part) => part[0] !== '.');
 
         if (
-          (hasAnotherComponentInPath(pathParts) && pathParts.length > 1) ||
-          (hasComponentDirectoryInPath(pathParts) && pathParts.length > 2)
+          isASupportedFileType(pathParts) &&
```
cartogram: `isASupportedFileType` => `isSupportedFileType`

cartogram (approved):

cartogram:
@alexandcote LGTM, did you want to do a patch release?

lib/rules/strict-component-boundaries.js
cartogram: On second thought, can you add `.graphql`

lemonmade:
IMO excluding the fixtures folder is more appropriate here. Some components could easily have assets we don’t want people reaching for (i.e., our emoji picker has a json file for th emoji definition; components have their own private sass files, etc)

cartogram:
@lemonmade okay, that makes sense to me... didn't realize there were other places where we would be storing `json` next to a component.

alexandcote:
@lemonmade / @cartogram 👀

lib/rules/strict-component-boundaries.js
```diff
@@ -39,3 +40,7 @@ function hasComponentDirectoryInPath(pathParts) {
 function hasAnotherComponentInPath(pathParts) {
   return Boolean(pathParts.filter((part) => part === pascalCase(part)).length);
 }
+
+function isNotAFixtureDirectory(pathParts) {
+  return Boolean(pathParts.filter((part) => part === 'fixtures').length === 0);
```
lemonmade: No need for the `Boolean()`

tests/lib/rules/strict-component-boundaries.js
```diff
@@ -25,6 +25,10 @@ ruleTester.run('strict-component-boundaries', rule, {
       code: `import {someThing} from '../OtherComponent';`,
       parserOptions,
     },
+    {
+      code: `import someThing from '../OtherComponent/fixtures/mock-query.json';`,
```
lemonmade: I still think this should be caught, but `../fixtures/OtherComponent/foo.json` should not (this means reaching up into another component, and getting its fixture, which doesn't seem right).

lemonmade (changes requested):

alexandcote:
Ok, I dig more into this and I found a lot of issue with this rules.

1. We should probably be able to include everything from an external library like:
```
import {Subscription} from 'apollo-client/util/Observable';
```
This trigger a lint error because `Observable` is PascalCase.

2. You may need a folder in the GraphQL folder to structure your queries: 
```
import something from '../graphql/SomeMockQuery/Query.graphql'
```

3. Tests can need child component:
```
import MyComponent from '../components/MyComponent'
import MyComponent from '../components/MyComponent/MyComponent'
```

4. You may need types from another component : 
```
import {MyType} from 'components/MyComponent/types'
import {MyType} from '../MyComponent/types'
```

5. You may need types from a graphql query of another component : 
```
import {MyType} from 'components/MyComponent/types'
```

6. You may prefer to use the absolute path:
```
sections/MyComponent/MyComponent.tsx

import {MyComponent} from 'sections/MyComponent/components'
```

I only list some of them. I think it will be really hard to handle all those case.
So my question is what we really want to prevent with `strict-component-boundaries`

cartogram:
@alexandcote We want to prevent all of those except the first one.

lemonmade:
We should actually be able to solve the first one, too. Using `resolve` from the `eslint-module-utils` package to get the fully resolved path for a module (see [example](https://github.com/Shopify/eslint-plugin-shopify/blob/master/lib/utilities.js#L42)). That will let us tell if it is in node modules or part of the real project.

cartogram:
@alexandcote let me know if you need help wrapping this up.

alexandcote:
Ready for 👀

cartogram:
@alexandcote additions LGTM. 🚢

PR #121, opened 2018-07-27
Remove inline Prettier configs (fixes #118)

### LINKED ISSUE #118: Prettier configs override prettierrc
`plugin:shopify/typescript-prettier` and `plugin:shopify/prettier` pass Prettier config inline overriding any external config. It can be worked around but it took me a few hours to figure out why:

```yaml
  prettier/prettier:
    - error
    - parser: typescript
```

texastoland:
Theoretically no breaking changes. The TypeScript parser upgrade fixes deprecated API calls broken for TypeScript 3.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixed
+* Updated `typescript-eslint-parser` dependency to version 17.0.1 in order to support TypeScript 3.
+* `plugin:shopify/prettier` and `plugin:shopify/typescript-prettier` now defer Prettier's config to project's `.prettierrc`.
+
 ### Changed
 * Included `all` as a vague term for `no-vague-titles` ([#114](https://github.com/Shopify/eslint-plugin-shopify/pull/114))
 
--- a/.prettierrc
+++ b/.prettierrc
@@ -2,5 +2,6 @@
   "singleQuote": true,
   "bracketSpacing": false,
   "trailingComma": "all",
+  "jsxBracketSameLine": false,
   "arrowParens": "always"
 }
--- a/lib/config/prettier.js
+++ b/lib/config/prettier.js
@@ -4,17 +4,6 @@ module.exports = {
   plugins: ['prettier'],
 
   rules: {
-    'prettier/prettier': [
-      'error',
-      {
-        singleQuote: true,
-        trailingComma: 'all',
-        bracketSpacing: false,
-        jsxBracketSameLine: false,
-        arrowParens: 'always',
-      },
-    ],
-
     // rules to disable to prefer prettier
     'shopify/binary-assignment-parens': 'off',
     'babel/semi': 'off',
--- a/lib/config/typescript-prettier.js
+++ b/lib/config/typescript-prettier.js
@@ -6,10 +6,6 @@ module.exports = {
       'error',
       {
         parser: 'typescript',
-        singleQuote: true,
-        trailingComma: 'all',
-        bracketSpacing: false,
-        jsxBracketSameLine: false,
       },
     ],
   },
--- a/package.json
+++ b/package.json
@@ -99,6 +99,6 @@
     "merge": "1.2.0",
     "pascal-case": "^2.0.1",
     "pkg-dir": "2.0.0",
-    "typescript-eslint-parser": "16.0.1"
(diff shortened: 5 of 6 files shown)


GoodForOneFare (approved):
LGTM 👍 Tested in a large internal project with no issues.

@ismail-syed - are you cool with this?  I think the original intent was to provide a baseline for all Shopify projects, but it looks like that introduces inflexibility.

cc @cartogram - does webgen provide a default .prettierrc?  If not, it could be useful to add one if this change is merged.

texastoland:
🙌🏽 thank you guys your configs are bar none I should try Polaris!

ismail-syed:
The `typescript-eslint-parser` bump makes sense, but i'm still a bit confused behind the intent of removing the our `prettier/prettier` config. When consumers extend `plugin:shopify/typescript-prettier` or `"plugin:shopify/prettier"`, we want them to the defaults we've set. 
This prevents that now, right?

texastoland:
A linter plugin shouldn't override the formatter config. Prettier should be configured (as Shopify does) via Cosmiconfig (`.prettierrc` in your case or `prettierrc.yaml`/`package.json` in mine). Passing them inline via ESLint overrode all other configuration locations.

Aside it took hours to debug that an ESLint config (surprise) made my Prettier config stop working. I first checked the editor extension, then Prettier itself, and finally Prettier plugins before checking `plugin:shopify/typescript-prettier`'s implementation.

`jsxBracketSameLine` was the only setting not mirrored in `.prettierrc` so I added it to prevent breaking Shopify's projects.

ismail-syed:
Our setup is like that because we run prettier against our scripts files (js, ts, tsx) as an eslint plugin. This provides us with a better integration of prettier with or existing eslint config and better error display. AFAIK, the standalone prettier cli only displays which files are being flagged. ESlint provides a better line by line error output. 

That being said, we have had our fair share of config debugging. Do you have a small repo I can look at to see where our config stops your prettier config from not working?

texastoland:
@ismail-syed this doesn't change your use case in any way. Running Prettier via ESLint still uses your `.prettierrc`. All inline options do is override user configs. PS I want your configs precisely because they're the strictest, preferential to TypeScript, and integrate nicely with Prettier.

It's easy to repro in your own workspace. Edit `.prettierrc` in _any_ project to `"singleQuote": false` (single quotes are inconsistent with other language paradigms). The change will be ignored by Prettier. Same for `arrowParens`, `bracketSpacing`, or `trailingComma`.

ismail-syed (approved):
@texastoland, that's for clarification and contribution! Was able to confirm everything you mentioned was correct.  

One comment, can you please add a note about this in the changelog.md.

texastoland:
Will do thank you for your consideration @ismail-syed!

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixed
+* Updated `typescript-eslint-parser` dependency to version 17.0.1 in order to support TypeScript 3.
```
ismail-syed, line 6: Can this be moved to the `Changed` section

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
...
+* Removed Prettier configs from `plugin:shopify/prettier` and `plugin:shopify/typescript-prettier` and added missing `jsxBracketSameLine` rule to `.prettierrc`.
```
ismail-syed: > `plugin:shopify/prettier` and `plugin:shopify/typescript-prettier` configs now defer prettier configs to project's `.prettierrc`

texastoland: 👍🏽 amended.

CHANGELOG.md
ismail-syed: Is this back words compatible with previous versions of ts?

texastoland: 💯 the bump is backwards-compatible see eslint/typescript-eslint-parser#499. Without it ESLint crashes hard with the TypeScript 3 compiler. I consider it a fix but will move it if you prefer.

🤷🏽‍♂️ as an aside the parser itself is not backwards-compatible in any way to breaking changes. I recommended a fix but it was ignored. It doesn't even use types.

ismail-syed:
Thanks again for addressing all the feedback @texastoland. Gonna go ahead and merge this.

PR #133, opened 2018-08-03
Adds `prefer-name-spaced-react’

### LINKED ISSUE #115: Prefer named space import React
```
// incorrect
import React from 'react'

// correct
import * as React from 'react'
```

cartogram:
closes https://github.com/Shopify/web-foundation/issues/115

If I can get approval on the general approach and if we want this rule, I'll fill in the docs, configs, changelog, etc...

### DIFF
--- a/docs/rules/prefer-name-spaced-react.md
+++ b/docs/rules/prefer-name-spaced-react.md
@@ -1,0 +1,36 @@
+# Prefer React import be name-spaced. (prefer-name-spaced-react)
+
+Please describe the origin of the rule here.
+
+
+## Rule Details
+
+This rule aims to...
+
+Examples of **incorrect** code for this rule:
+
+```js
+
+// fill me in
+
+```
+
+Examples of **correct** code for this rule:
+
+```js
+
+// fill me in
+
+```
+
+### Options
+
+If there are any options, describe them here. Otherwise, delete this section.
+
+## When Not To Use It
+
+Give a short description of when it would be appropriate to turn off this rule.
+
+## Further Reading
+
+If there are other links that describe the issue this rule addresses, please include them here in a bulleted list.
--- a/lib/rules/prefer-name-spaced-react.js
+++ b/lib/rules/prefer-name-spaced-react.js
@@ -1,0 +1,34 @@
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Prefer React import be name-spaced.',
+      category: 'Stylistic Issues',
+      recommended: false,
+      uri:
+        'https://github.com/Shopify/eslint-plugin-shopify/blob/master/docs/rules/prefer-name-spaced-react.md',
+    },
+    fixable: null,
+  },
+  create(context) {
+    function report(node) {
+      const {name} = node;
+
+      context.report({
+        node,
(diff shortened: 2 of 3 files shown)


lib/rules/prefer-name-spaced-react.js
```diff
@@ -0,0 +1,34 @@
...
+        message: `Prefer name-spaced React import. (import * as React from 'react';).`,
+        data: {name},
+      });
+    }
+
+    return {
+      ImportDeclaration(node) {
+        if (
+          node.source.value === 'react' &&
+          node.specifiers[0].type === 'ImportDefaultSpecifier'
```
lemonmade, line 27: You probably need to guard against no specifiers

lemonmade (commented):
Approach looks good 👍

GoodForOneFare:
Is just forbidding `allowSyntheticDefaultImports` an option?

cartogram:
I might prefer the rule over `allowSyntheticDefaultImports`, @lemonmade?

Also, would this be under the typescript namespace as well?

lemonmade:
I think this can be applied in JS or TS, it works the same either way.

Disallowing `allowSyntheticDefaultImports` could work but it is bad for other modules like `express`, so I would prefer to avoid

docs/rules/prefer-name-spaced-react.md
```diff
@@ -0,0 +1,36 @@
+# Prefer React import be name-spaced. (prefer-name-spaced-react)
```
TzviPM, line 1: did you mean to fill this in?

lib/rules/prefer-name-spaced-react.js
```diff
@@ -0,0 +1,34 @@
...
+    },
+    fixable: null,
+  },
+  create(context) {
+    function report(node) {
+      const {name} = node;
+
+      context.report({
+        node,
+        message: `Prefer name-spaced React import. (import * as React from 'react';).`,
```
TzviPM, line 18: I think we actually want the opposite here.

TzviPM (changes requested):
I believe we actually want the opposite behaviour here as per https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html:

> We highly recommend applying it (`esModuleInterop`) both to new and existing projects. For existing projects, namespace imports (import * as express from "express"; express();) will need to be converted to default imports (import express from "express"; express();).

lib/rules/prefer-name-spaced-react.js
lemonmade: I am curious whether TS will complain about the namespace import for React with that flag applied. My gut is that a namespace import is still more correct for React. In express, it made a lot less sense, because the namespace import got you a function, where it would always give you an object in “true” modules. React, in contrast, uses module.exports as an object that just exposes named properties, so a namespace import is more appropriate.

docs/rules/prefer-name-spaced-react.md
cartogram: oh yes, sorry, I forgot i was just looking for general approach approval at first. Will clean this all up.

cartogram:
superseded  by https://github.com/Shopify/eslint-plugin-shopify/pull/262

Issue #127, opened 2018-08-01
Add a rule to prefer no `renderX` methods in React components


lemonmade:
Usually better to do a component or just fold into `render`

PR #134, opened 2018-08-03
adds `react-no-render-methods’ rule

cartogram:
Closes https://github.com/Shopify/eslint-plugin-shopify/issues/127

Looking for feedback on the name and general approach.


lib/rules/react-no-render-methods.js
```diff
@@ -0,0 +1,46 @@
+const Components = require('eslint-plugin-react/lib/util/Components');
```
lemonmade: Name is a little misleading since there is still a render method 🙃

lib/rules/react-no-render-methods.js
```diff
@@ -0,0 +1,46 @@
...
+    let isES6Component = false;
+
+    function report(node) {
+      const {
+        key: {name},
+      } = node;
+
+      context.report({
+        node,
+        message: `No renderX methods. '{{name}}' should be a seperate component.`,
```
lemonmade: Message is a little too direct

lemonmade (commented):
Looks good!

lib/rules/react-no-render-methods.js
cartogram: what about `react-no-multiple-render-methods`?

marutypes (approved):
I love this, great explanation 👏

PR #135, opened 2018-08-03
Fix semicolon rules conflicting with Prettier

texastoland:
I believe the only changes in behavior legitimately didn't work before. The changes to Flow were copied from TypeScript. The changes for Prettier could be checked by an integration test with a `.prettierrc` configured opposite of Shopify's.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # Changelog
 
 <!-- ## Unreleased -->
+### Fixed
+* `plugin:shopify/flow` now disables rules checked by Flow's static analyzer. ([#135](https://github.com/Shopify/eslint-plugin-shopify/pull/135))
+* `plugin:shopify/prettier` and `plugin:shopify/typescript-prettier` defer missing semicolon rules to a project´s `.prettierrc`. ([#135](https://github.com/Shopify/eslint-plugin-shopify/pull/135))
+
 ### Added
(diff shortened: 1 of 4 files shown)


texastoland:
May I recommend an integration test with a `.prettierrc` whose options are configured opposite of Shopify's? That would catch this and #118 in case of regression or new rules.

texastoland:
@ismail-syed ready for review in case you want an integration test.

lib/config/rules/typescript.js
```diff
@@ -101,7 +101,6 @@ module.exports = {
   'no-useless-constructor': 'off',
   'no-shadow': 'off',
   'no-use-before-define': 'off',
-  'sort-class-members/sort-class-members': 'off',
```
ismail-syed: Why is this being removed? 
For typescript, we enforce member ordering with [`typescript/member-ordering`](https://github.com/Shopify/eslint-plugin-shopify/blob/master/lib/config/rules/typescript.js#L48-L67), not eslint's equivalent.

texastoland: 👍🏽 ah I totally missed that. I'll check it works and amend.

texastoland: 👌🏽 fixed and rebased.

ismail-syed (approved):
👍 Thanks!

PR #139, opened 2018-08-03
adds rule to enforce pascal case on test IDs

kvendrik:


kvendrik:
Closing this as `testID`s are now [deprecated](https://github.com/Shopify/web-foundation/blob/master/Decision%20records/04%20-%20We%20do%20not%20use%20test%20IDs.md).

PR #140, opened 2018-08-03
[strict-component-boundaries] Exclude node_modues

cartogram:
@lemonmade took a stab at this one, let me know what you think of the approach and how I would go about testing this. 

The issue I am having is that in order for the `inNodeModules` function to work, the package needs to actually be in `node_modules` and to test that I need to have a module installed that breaks the `strict-component-imports` rule, which I don't. 

For example, I would need to have `import {Subscription} from 'apollo-client/util/Observable';` installed to this project.

cc/ @TheMallen


lib/rules/strict-component-boundaries.js
```diff
@@ -32,6 +38,20 @@ module.exports = {
...
 };
 
+function inNodeModules(importSource, context) {
+  const resolvedSource = resolve(importSource, context);
+
+  if (resolvedSource === null) {
+    return true;
+  }
+
+  if (!resolvedSource) {
```
lemonmade: What do the two checks above do?

lib/rules/strict-component-boundaries.js
```diff
@@ -32,6 +38,20 @@ module.exports = {
...
+    return false;
+  }
+
+  return resolvedSource.match(/node_modules/);
```
lemonmade: Usually better to use regex.test since it returns a real Boolean

tests/lib/rules/strict-component-boundaries.js
```diff
@@ -25,6 +25,14 @@ ruleTester.run('strict-component-boundaries', rule, {
       code: `import {someThing} from '../OtherComponent';`,
       parserOptions,
     },
+    {
+      code: `import fs from 'fs';`,
```
cartogram: these 2 added tests are not right, they should have something that would otherwise trigger this rule, but because they are in node modules (or a core package) they would be valid.

tests/lib/rules/strict-component-boundaries.js
```diff
@@ -25,6 +25,14 @@ ruleTester.run('strict-component-boundaries', rule, {
...
+      parserOptions,
+    },
+    {
+      code: `import eslint from 'eslint';`,
+      parserOptions,
+    },
```
lemonmade: Would be good to test that this actually works for a path that is in node modules and that does not adhere to our rules (something/MyComponent)

lemonmade (changes requested):

lib/rules/strict-component-boundaries.js
cartogram: `resolvedSource  === null` if it is a core library
`resolvedSource === undefined` if it is not installed

finally if we get a string, its the full path to the module.

tests/lib/rules/strict-component-boundaries.js
cartogram: see https://github.com/Shopify/eslint-plugin-shopify/pull/140#discussion_r207691748

cartogram: I need to actually install `something/MyComponent`, otherwise I don't think this will work.

lemonmade: See example here: 
https://github.com/Shopify/eslint-plugin-shopify/blob/master/tests/lib/rules/jsx-no-hardcoded-content.js#L156

cartogram:
@lemonmade cleaned things up and added the node_modules test.

tests/lib/rules/strict-component-boundaries.js
cartogram: 🙇

lib/rules/strict-component-boundaries.js
```diff
@@ -32,10 +39,22 @@ module.exports = {
   },
 };
 
+function isCoreModule(resolvedSource) {
```
alexandcote, line 42: What is a `CoreModule` ?

tests/lib/rules/strict-component-boundaries.js
```diff
@@ -25,6 +26,11 @@ ruleTester.run('strict-component-boundaries', rule, {
       code: `import {someThing} from '../OtherComponent';`,
       parserOptions,
     },
+    {
+      code: `import {getDisplayName} from '@shopify/react-utilities/components';`,
+      parserOptions,
+      filename: fixtureFile('basic-app/app/sections/MySection/MySection.js'),
```
alexandcote, line 32: What it does `fixtureFile` ?

alexandcote (approved):

lib/rules/strict-component-boundaries.js
cartogram: A module that comes with node core

Issue #101, opened 2018-06-24
Support `describe.each` syntax in no-vague-titles


cartogram:
https://github.com/Shopify/eslint-plugin-shopify/pull/93#discussion_r191204727

PR #148, opened 2018-08-28
Adds support for describe.each to no-vague-titles

cartogram:
Closes https://github.com/Shopify/eslint-plugin-shopify/issues/101

This was suddenly more urgent because sewing-kit started writing tests in this style. The lack of support for this feature was causing errors when linting that codebase.

### DIFF
--- a/lib/rules/jest/no-vague-titles.js
+++ b/lib/rules/jest/no-vague-titles.js
@@ -75,15 +75,18 @@ function hasEmptyDescription({arguments: args}) {
 }
 
 function getMethodName({callee}) {
-  if (callee.name) {
-    return callee.name;
+  switch (callee.type) {
+    case 'CallExpression':
+      return callee.callee.object.name;
+    case 'Identifier':
+      return callee.name;
+    case 'MemberExpression':
+      return callee.object.name;
+    case 'ArrowFunctionExpression':
+      return '';
+    default:
+      throw new Error('Could not get method name from node.');
   }
-
-  if (callee.object) {
-    return callee.object.name;
-  }
-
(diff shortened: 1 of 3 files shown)


lib/rules/jest/no-vague-titles.js
```diff
@@ -79,6 +74,18 @@ function hasEmptyDescription({arguments: args}) {
...
+function getMethodName({callee}) {
+  if (callee.name) {
+    return callee.name;
+  }
+
+  if (callee.object) {
+    return callee.object.name;
+  }
+
+  return callee.type === 'CallExpression' && callee.callee.object.name;
```
lemonmade: A switch on the type is probably cleaner here (and maybe throw in the unknown case)

tests/lib/rules/jest/no-vague-titles.js
```diff
@@ -593,11 +630,15 @@ ruleTester.run('no-tests-contain-correct with ignore=describe', rule, {
       code: "describe.only('appropriate')",
       options: [{ignore: ['describe']}],
     },
+    {
+      code: `describe.each([['production'], ['staging']])('Includes things all for %s clients')`,
```
lemonmade: Doesn’t include any of the blacklisted words?

lemonmade: Oh, I guess all is restricted, seems weird. In any case, the valid and invalid examples are much too subtle imo.

lemonmade (approved):
A few minor suggestions but looks good

ismail-syed (approved):

Issue #138, opened 2018-08-03
consistent/explicit imports to relative/local components


cartogram:
A common pattern found in test files. The `index.ts` should be reserved for exposing pieces of the component to the outside application while test files themselves should import from the individual file they are testing.

```
// bad
import {Button} from '..';

// good
import {Button} from '../Button';
```

PR #149, opened 2018-08-28
Adds `no-ancestor-directory-import` rule

cartogram:
Closes https://github.com/Shopify/eslint-plugin-shopify/issues/138

In addition to the code, feedback on the description and name of this rule would be really great. I found it to be a hard one to describe in words.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 <!-- ## Unreleased -->
 
+### Added
+* `shopify/no-ancestor-directory-import` ([#149](https://github.com/Shopify/eslint-plugin-shopify/pull/149))
+
 ## [26.0.0] - 2018-10-26
 
 ### Added
--- a/docs/rules/no-ancestor-directory-import.md
+++ b/docs/rules/no-ancestor-directory-import.md
@@ -1,0 +1,27 @@
+# Prefer that imports from within a directory extend to the file from where they are importing without relying on an index file. (no-ancestor-directory-import)
+
+Imports inside the same directory should extend directly to the file from where they are importing without relying on an index file. This means the source of these imports should not end with a directory (`/`), but the path should terminate at an individual filename. This preserves the index file inside a directory as a mechanism for exposing pieces of the module to the outside application, rather than as a way to export the parts that the module internally depends on.
+
+## Rule Details
+
+This rule disallows any full directory imports from within that same directory.
+
+Examples of **incorrect** code for this rule:
+
+```ts
+import Thing from '../';
+import OtherThing from './';
+import Module from '../../index.ts'
+```
+
+Examples of **correct** code for this rule:
+
+```ts
+import Thing from '../Thing';
+import OtherThing from './OtherThing';
+import Module from '../Module'
+```
+
+## When Not To Use It
+
+If you do not wish to prevent directory imports from within that directory, you can safely disable this rule.
--- a/lib/rules/no-ancestor-directory-import.js
+++ b/lib/rules/no-ancestor-directory-import.js
@@ -1,0 +1,49 @@
+const {extname, basename, relative, sep} = require('path');
+const resolve = require('eslint-module-utils/resolve').default;
+
+module.exports = {
+  meta: {
+    docs: {
+      description:
+        'Prefer that imports from within a directory extend to the file from where they are importing without relying on an index file.',
+      category: 'Best Practices',
+      recommended: false,
+      uri:
+        'https://github.com/Shopify/eslint-plugin-shopify/blob/master/docs/rules/no-ancestor-directory-import.md',
+    },
+    fixable: null,
+  },
+  create(context) {
+    function isAncestorDirectoryImport(resolvedSource) {
+      const relativeDifference = relative(
+        context.getFilename(),
+        resolvedSource,
+      );
+
+      const parts = relativeDifference
+        .split(sep)
+        .filter((part) => part[0] !== '.');
+
+      if (parts.length > 1) {
+        return false;
+      }
+
+      return basename(parts[0], extname(parts[0])) === 'index';
+    }
+
+    return {
+      ImportDeclaration(node) {
+        const importSource = node.source.value;
+        const resolvedSource = resolve(importSource, context);
+
+        if (resolvedSource && isAncestorDirectoryImport(resolvedSource)) {
+          context.report({
+            node,
+            message:
+              'Ancestor imports should extend to the file from where they are importing without relying on an index file in the directory.',
+          });
+        }
+      },
+    };
+  },
+};
--- a/tests/lib/rules/no-ancestor-directory-import.js
+++ b/tests/lib/rules/no-ancestor-directory-import.js
@@ -1,0 +1,105 @@
+const {RuleTester} = require('eslint');
+const {fixtureFile} = require('../../utilities');
+const rule = require('../../../lib/rules/no-ancestor-directory-import');
+
+const ruleTester = new RuleTester({
+  parserOptions: {
+    ecmaVersion: 6,
+    sourceType: 'module',
+  },
+});
+
+const errors = [
+  {
+    type: 'ImportDeclaration',
+    message:
+      'Ancestor imports should extend to the file from where they are importing without relying on an index file in the directory.',
+  },
+];
+
+ruleTester.run('no-ancestor-directory-import', rule, {
+  valid: [
+    {
+      code: "import Foo from '../Foo'",
+      filename: fixtureFile('basic-app/app/components/Foo/tests/Foo.test.js'),
+    },
+    {
+      code: "import {Foo, Bar} from '../../components'",
+      filename: fixtureFile('basic-app/app/sections/MySection/MySection.js'),
+    },
+    {
+      code: "import Bar from '../Bar'",
+      filename: fixtureFile('basic-app/app/components/Foo/Foo.js'),
+    },
+    {
+      code: "import Bar from '../Bar'",
+      filename: fixtureFile(
+        'basic-app/app/components/Foo/components/Baz/Baz.js',
+      ),
+    },
+    {
+      code: "import Qux from '../Qux'",
+      filename: fixtureFile(
+        'basic-app/app/components/Foo/components/Baz/Baz.js',
+      ),
+    },
+    {
+      code: "import Bar from '../../../Bar'",
+      filename: fixtureFile(
+        'basic-app/app/components/Foo/components/Baz/Baz.js',
+      ),
+    },
+  ],
+
+  invalid: [
+    {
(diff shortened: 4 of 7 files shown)


docs/rules/prefer-explicit-local-imports.md
```diff
@@ -0,0 +1,27 @@
+# Disallow vague relative local imports. (prefer-explicit-local-imports)
+
+The index file inside a directory should be reserved for exposing pieces of the module to the outside application, while the files inside that module should import individual files directly.
+
+## Rule Details
+
+This rule aims to...
```
kaelig: Is this sentence incomplete?

cartogram: @kaelig updated

docs/rules/prefer-explicit-local-imports.md
```diff
@@ -0,0 +1,27 @@
+# Prefer relative local imports extend to the file from where they are importing without relying on an index file. (prefer-explicit-local-imports)
```
lemonmade: I think I might describe this more as "no ancestor directory import".

lib/rules/prefer-explicit-local-imports.js
```diff
@@ -0,0 +1,39 @@
...
+            message:
+              'Relative local imports should extend to the file from where they are importing without relying on an index file.',
+          });
+        }
+      },
+    };
+  },
+};
+
+function isVagueImport(str) {
```
lemonmade: It isn't really vague, it's more about the fact that it is a direct ancestor.

lib/rules/prefer-explicit-local-imports.js
```diff
@@ -0,0 +1,39 @@
...
+      recommended: false,
+      uri:
+        'https://github.com/Shopify/eslint-plugin-shopify/blob/master/docs/rules/prefer-explicit-local-imports.md',
+    },
+    fixable: null,
+  },
+  create(context) {
+    return {
+      ImportDeclaration(node) {
+        const pathParts = node.source.value.split('/');
```
lemonmade: Use `path.sep`

lib/rules/prefer-explicit-local-imports.js
```diff
@@ -0,0 +1,39 @@
...
+    };
+  },
+};
+
+function isVagueImport(str) {
+  if (str.length <= 2) {
+    return str === '' || str === '.' || str === '..';
+  }
+
+  return basename(str, extname(str)) === 'index';
```
lemonmade: I think you are better off resolving the path and then checking that either it is a directory and an ancestor of the current path, or it is an index file of such a directory. This prevents potential issues with this method, as this incorrectly warns on some weirder import paths that we'd still consider technically valid

lemonmade (changes requested):

lib/rules/prefer-explicit-local-imports.js
ismail-syed: >  some weirder import paths that we'd still consider technically valid

What would some of these cases be? It would be good to add them to the tests

tests/lib/rules/prefer-explicit-local-imports.js
```diff
@@ -0,0 +1,63 @@
...
+
+ruleTester.run('prefer-explicit-local-imports', rule, {
+  valid: [
+    {
+      code: "import Button from '../Button'",
+      parserOptions,
+    },
+  ],
+
+  invalid: [
```
ismail-syed: Maybe add a few valid and invalid tests for importing non-default exports
eg. `import {Foo, Bar} from '../'`

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 ### Added
 * `shopify/prefer-singular-enums` ([#132](https://github.com/Shopify/eslint-plugin-shopify/pull/132))
 * `shopify/react-no-multiple-render-methods` ([#134](https://github.com/Shopify/eslint-plugin-shopify/pull/134))
+* `shopify/prefer-explicit-local-imports` ([#149](https://github.com/Shopify/eslint-plugin-shopify/pull/149))
```
ismail-syed: `shopify/prefer-namespaced-imports` might be a better name - not entirely sure.

ismail-syed (approved):
👍 Nice one. Nothing major from my end, just a couple comments.

docs/rules/prefer-explicit-local-imports.md
cartogram: Yeah, all makes sense. I like that name.

cartogram:
@lemonmade ready for another look. Thanks!

cc/ @BPScott since you ran into this recently.

tests/lib/rules/no-ancestor-directory-import.js
```diff
@@ -0,0 +1,71 @@
+const {RuleTester} = require('eslint');
+const {fixtureFile} = require('../../utilities');
+const rule = require('../../../lib/rules/no-ancestor-directory-import');
+
+const ruleTester = new RuleTester();
```
BPScott: I think you can do 

```js
const parserOptions =
new RuleTester({parserOptions:  {
  ecmaVersion: 6,
  sourceType: 'module',
}})
```

to avoid having to pass parserOptions into every valid/invalid block below. (see https://eslint.org/docs/developer-guide/nodejs-api#ruletester)

BPScott (commented):
Nice. A few more test cases around components reaching into sibling components, and subcomponents would be handy. If I've understood the purpose of this PR then the following should all pass:

```js
{
    code: "import Bar from '../Bar'",
    filename: fixtureFile('basic-app/app/components/Foo/Foo.js'),
},
{
    code: "import Bar from '..'",
    filename: fixtureFile('basic-app/app/components/Foo/Foo.js'),
},
{
    code: "import Bar from '../Bar'",
    filename: fixtureFile('basic-app/app/components/Foo/components/Baz/Baz.js'),
},
{
    code: "import Qux from '..'",
    filename: fixtureFile('basic-app/app/components/Foo/components/Baz/Baz.js'),
},
{
    code: "import Qux from '../Qux'",
    filename: fixtureFile('basic-app/app/components/Foo/components/Baz/Baz.js'),
},
{
    code: "import Bar from '../../../Bar'",
    filename: fixtureFile('basic-app/app/components/Foo/components/Baz/Baz.js'),
},
{
    code: "import Bar from '../../..'",
    filename: fixtureFile('basic-app/app/components/Foo/components/Baz/Baz.js'),
},
```

docs/rules/prefer-explicit-local-imports.md
```diff
@@ -0,0 +1,27 @@
+# Prefer imports from within a directory extend to the file from where they are importing without relying on an index file. (no-ancestor-directory-import)
```
lemonmade: `Prefer that`

lib/rules/no-ancestor-directory-import.js
```diff
@@ -0,0 +1,49 @@
+const {extname, basename, relative, sep} = require('path');
+const resolve = require('eslint-module-utils/resolve').default;
+
+module.exports = {
+  meta: {
+    docs: {
+      description:
+        'Prefer imports from within a directory extend to the file from where they are importing without relying on an index file.',
```
lemonmade: Here too

lib/rules/no-ancestor-directory-import.js
```diff
@@ -0,0 +1,49 @@
...
+      category: 'Best Practises',
```
lemonmade: `Best practices`

lib/rules/no-ancestor-directory-import.js
```diff
@@ -0,0 +1,49 @@
...
+      recommended: false,
+      uri:
+        'https://github.com/Shopify/eslint-plugin-shopify/blob/master/docs/rules/no-ancestor-directory-import.md',
+    },
+    fixable: null,
+  },
+  create(context) {
+    function isAncestorDirectoryImport(resolvedSource) {
```
lemonmade, line 17: This function doesn't seem to quite line up with what it checks

lemonmade (approved):

tests/lib/rules/no-ancestor-directory-import.js
cartogram: TIL! Thanks @BPScott

docs/rules/no-ancestor-directory-import.md
```diff
@@ -0,0 +1,27 @@
+# Prefer that imports from within a directory extend to the file from where they are importing without relying on an index file. (no-ancestor-directory-import)
+
+Imports inside the same directory should extend directly to the file from where they are importing without relying on an index file. This means the source of these imports should not end with a directory (`/`), but the path should terminate at an individual filename. This preserves the index file inside a directory as a mechanism for exposing pieces of the module to the outside application.
```
lemonmade: The last sentence should be augmented with `, rather than as a way to export the parts that the module internally depends on.` or something like that.

lib/rules/no-ancestor-directory-import.js
```diff
@@ -0,0 +1,49 @@
...
+    return {
+      ImportDeclaration(node) {
+        const importSource = node.source.value;
+        const resolvedSource = resolve(importSource, context);
+
+        if (resolvedSource && isAncestorDirectoryImport(resolvedSource)) {
+          context.report({
+            node,
+            message:
+              'Relative local imports should extend to the file from where they are importing without relying on an index file.',
```
lemonmade: I don't know if this message is quite right; since we resolve the source, it might not even be a relative import (i.e., something in `app/sections` asking for `index` would fail this rule, as that likely resolves to `app/index.js`

lib/rules/no-ancestor-directory-import.js
```diff
@@ -0,0 +1,49 @@
...
+    fixable: null,
+  },
+  create(context) {
+    function isAncestorDirectoryImport(resolvedSource) {
+      const relativeDifference = relative(
+        context.getFilename(),
+        resolvedSource,
+      );
+
+      const parts = relativeDifference
```
lemonmade, line 23: So this just strips out all the leading `./../../` stuff right?

lemonmade (approved):

tests/lib/rules/no-ancestor-directory-import.js
```diff
@@ -0,0 +1,98 @@
...
+      errors,
+      filename: fixtureFile('basic-app/app/components/Foo/tests/Foo.test.js'),
+    },
+    {
+      code: "import Bar from '..'",
+      errors,
+      filename: fixtureFile('basic-app/app/components/Foo/Foo.js'),
+    },
+    {
+      code: "import Bar from '../../..'",
```
lemonmade, line 91: Might want to check that `../../../` works also (trailing slash)

PR #150, opened 2018-08-28
Fix ignore `fix-strict-component-boundaries` fixtures folder

cartogram:
This PR simplifies the fixtures folder check in `strict-component-boundaries`.  In this PR we just check if we are looking inside a fixtures directory and bail if we are. IMO thats all we need.

### DIFF
--- a/tests/lib/rules/strict-component-boundaries.js
+++ b/tests/lib/rules/strict-component-boundaries.js
@@ -57,15 +57,5 @@ ruleTester.run('strict-component-boundaries', rule, {
       parserOptions,
       errors,
     },
-    {
-      code: `import someThing from '../SomeComponent/fixtures/SomeMockQuery/query.json';`,
-      parserOptions,
-      errors,
-    },
-    {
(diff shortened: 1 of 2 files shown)


tests/lib/rules/strict-component-boundaries.js
```diff
@@ -57,15 +57,5 @@ ruleTester.run('strict-component-boundaries', rule, {
       parserOptions,
       errors,
     },
-    {
-      code: `import someThing from '../SomeComponent/fixtures/SomeMockQuery/query.json';`,
```
lemonmade, line 61: So this import would now be valid? We certainly don't want it but maybe it just never comes up?

lemonmade (approved):

tests/lib/rules/strict-component-boundaries.js
cartogram: It would be and yeah I just don't think it comes up. I may come back to this.

Issue #123, opened 2018-07-31
Exclude node_modules from `strict-component-imports`


cartogram:
We can prevent the following from being flagged invalid by using the util identified in this discussion, https://github.com/Shopify/eslint-plugin-shopify/pull/117#issuecomment-408186647

```
import {Subscription} from 'apollo-client/util/Observable';
```

alexandcote:
This is not a bug anymore. Closing the issue.

Issue #122, opened 2018-07-28
strict-component-boundaries error message is vague


GoodForOneFare:
Violations just have a prosaic version of the rule name:
<img width="699" alt="index tsx neutron 2018-07-28 10-19-45" src="https://user-images.githubusercontent.com/673655/43357455-d86a2904-924f-11e8-9619-9898f301e6f0.png">

Users should not have to visit a rule's homepage to figure out a fix.

cartogram:
Agree. Couldn't think of a succinct way to express the error and/or fix. Any suggestions?

Also, that particular error would go away when I implement https://github.com/Shopify/eslint-plugin-shopify/issues/123

cartogram:
Closes via https://github.com/Shopify/eslint-plugin-shopify/pull/208

Issue #119, opened 2018-07-20
Should strict-component-boundaries check the current filename 


BPScott:
Occasionally components need to import their sibling components

E.g. in `src/components/Banner/Banner.ts`: importing a Button component `import Button from './Button';`

Currently this does not flag up an error. Should this be considered a boundary violation to make people change to use `import {Button} from '..'`;

//cc @cartogram

cartogram:
@BPScott why do you prefer `import {Button} from '..'`?

GoodForOneFare:
Unfortunate technical constraint: preferring index imports within a component hierarchy will eventually cause very, very hard to troubleshoot dependency ordering problems.

BPScott:
It seemed odd to me that in `src/components/Banner/Banner.ts` you could import the same module in two ways and one would pass the linting rule and one would fail:

```ts
// This passes strict-component-boundaries linting
import Button from '../Button'; 

// This imports exactly the same file, but would fail strict-component-boundaries linting
import Button from '../../components/Button';
```

As the fix for the second line failing in any file outside a components directory would be "change `import Button from 'some/path/components/Button';` to `import {Button} from 'some/path/components';`" I assumed the same would be true within the the components folder and that an import from the base would be the preferable approach.

It sounds like the first version is preferred if you're within a component folder thanks to Gord's technical constraints anyway.

Writing this up in the strict-component-boundaries docs would be good for modern-javascript-newbies like myself :)

cartogram:
@BPScott any additional documentation is welcome and appreciated if you want to open a PR, otherwise, it sounds like you got it and the rule is working as expected.

BPScott:
@GoodForOneFare makes it sound like there's a good reason why if you're in a "components" folder you should be using `import Button from './Button';` and not using `import {Button} from '..'`.

Currently `import {Button} from '..'` does not flag up an error if you're inside a "components". Should it throw a warning and suggest that you use `import Button from './Button';` instead?

Conversely `import Button from '../../components/Button';` does throw a warning if you're inside a "components" even though the @GoodForOneFare says that this is the correct behaviour - you should be hitting the component rather than the index in this case.

---

It sounds like strict-component-boundaries should still check the current path, but for the opposite reasons that I initially considered.

cartogram:
Yeah it could. Also feels like it could be a completely new rule.

PR #160, opened 2018-09-14
Fixes strict-component-boundaries

### LINKED ISSUE #156: strict-component-boundaries possibly flagging incorrectly
I can't tell if the `strict-component-boundaries` rule is working as intended in this case. These are utils, not components.

Here's the [errors in CI](https://circleci.com/gh/Shopify/images/29?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) for this [code](https://github.com/Shopify/images/pull/10/commits/d36cf94db2fafb60fc743ae1dbf9eaf884db039e). 

If this isn't a bug, please close this.
cc/ @cartogram

cartogram:
This PR resolves false positives with strict-component-boundaries that were introduced with https://github.com/Shopify/eslint-plugin-shopify/pull/140. 

Fixes https://github.com/Shopify/eslint-plugin-shopify/issues/156 https://github.com/Shopify/eslint-plugin-shopify/issues/122 and https://github.com/Shopify/eslint-plugin-shopify/issues/123

### Problem:
The check for whether there is a component in the path was returning true if the path to a module included a pascal case word. In https://github.com/Shopify/eslint-plugin-shopify/pull/140 this check changed from the ImportDeclaration source to the full path to the module on the users computer. Since `Shopify` is a pascal case word and in most cases there is at least one pascal case word in the full path as in `Users/your-username/path`, this was returning true in almost every ImportDeclaration.

### Fix:
The solution was to check the import source for the component in path check rather than the full path to the module source.

### 🎩 Instructions:

1. Pull this branch
1. Install yalc: `yarn add yalc -g`
1. run `yalc publish`
1. cd into a project that is currently reporting false positives for this rule (such as shopify/images)
1. install this branch using yalc: `yalc add eslint-plugin-shopify`
1. confirm that the false positives are no longer invalid

### DIFF
--- a/lib/rules/strict-component-boundaries.js
+++ b/lib/rules/strict-component-boundaries.js
@@ -17,19 +17,22 @@ module.exports = {
       ImportDeclaration(node) {
         const importSource = node.source.value;
         const resolvedSource = resolve(importSource, context);
-        const pathParts = resolvedSource
-          ? pathSegmantsFromSource(resolvedSource)
-          : pathSegmantsFromSource(importSource);
+        const pathParts = pathSegmantsFromSource(importSource);
 
-        if (isCoreModule(resolvedSource) || inNodeModules(pathParts)) {
+        if (
+          isCoreModule(resolvedSource) ||
+          inNodeModules(pathSegmantsFromSource(resolvedSource))
+        ) {
           return;
         }
 
         if (
-          ((hasAnotherComponentInPath(pathParts) && pathParts.length > 1) ||
-            (hasDirectoryInPath(pathParts, 'components') &&
-              pathParts.length > 2)) &&
-          !hasDirectoryInPath(pathParts, 'fixtures')
+          (hasAnotherComponentInPath(pathParts) &&
+            pathParts.length > 1 &&
+            !validFixtureImport(pathParts)) ||
+          (hasDirectoryInPath(pathParts, 'components') &&
+            pathParts.length > 2 &&
+            !validFixtureImport(pathParts))
         ) {
           context.report({
             node,
@@ -46,11 +49,29 @@ function isCoreModule(resolvedSource) {
 }
 
 function inNodeModules(pathParts) {
-  return Boolean(pathParts.filter((part) => part === 'node_modules').length);
+  return (
+    pathParts &&
+    Boolean(pathParts.filter((part) => part === 'node_modules').length)
+  );
 }
 
 function hasDirectoryInPath(pathParts, directory) {
   return Boolean(pathParts.filter((part) => part === directory).length);
+}
+
+function validFixtureImport(pathParts) {
+  if (!hasDirectoryInPath(pathParts, 'fixtures')) {
+    return false;
+  }
+
+  const fixtureIndexInPath = pathParts.findIndex((part) => part === 'fixtures');
+  const pathPartsBeforeFixture = pathParts.slice(0, fixtureIndexInPath);
+
+  if (!hasAnotherComponentInPath(pathPartsBeforeFixture)) {
+    return true;
+  }
+
+  return false;
 }
 
 function hasAnotherComponentInPath(pathParts) {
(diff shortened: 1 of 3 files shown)


BPScott:
I'm not going to pretend to know what this is doing under the hood but adding a bunch of examples to a test suite would make me feel better about it working

cartogram:
@BPScott thanks for pointing that out. I added what would be a test for this fix. However, this new test is pulling from a fixture folder and we are already ignoring when `fixtures` is in the path so the test will already never fail. 

I am going to need to go back and rework the `fixtures` ignore and update this PR. Changing the title to WIP in the mean time.

cartogram:
Updates:
1. solved fixture import so the added test is legit (it will fail without these changes)
1. successful 🎩 in `shopify/images` and `polaris-react`

tests/lib/rules/strict-component-boundaries.js
```diff
@@ -57,5 +62,10 @@ ruleTester.run('strict-component-boundaries', rule, {
       parserOptions,
       errors,
     },
+    {
+      code: `import someThing from '../SomeOtherComponent/fixtures/SomeMockQuery/query.json';`,
+      parserOptions,
+      errors,
+    },
```
lemonmade: I feel like you are not testing the main thing you indicated this PR was to address

lib/rules/strict-component-boundaries.js
```diff
@@ -17,19 +17,22 @@ module.exports = {
       ImportDeclaration(node) {
         const importSource = node.source.value;
         const resolvedSource = resolve(importSource, context);
-        const pathParts = resolvedSource
-          ? pathSegmantsFromSource(resolvedSource)
-          : pathSegmantsFromSource(importSource);
+        const pathParts = pathSegmantsFromSource(importSource);
```
lemonmade: In my opinion, this is not the most useful way to fix this issue. I would rather you never reference the image source, and instead operate on the relative source from the root of the app. Otherwise, I don’t think you’ll ever be able to get this to actually reflect what we want to enforce

lemonmade (changes requested):

tests/lib/rules/strict-component-boundaries.js
```diff
@@ -11,51 +11,62 @@ const parserOptions = {
 const errors = [
   {
     type: 'ImportDeclaration',
-    message: 'Strict component boundaries.',
+    message: `Do not reach into an individual component's folder for nested modules. Import from the closest shared components folder instead.`,
```
cartogram, line 14: cc @GoodForOneFare with regards to https://github.com/Shopify/eslint-plugin-shopify/issues/122

cartogram:
Update: 

- This rule now operates only on the full relative source. There are fewer edge cases when doing it this way. 
- This meant adding a "filename" property to each test that resolves to an actual fixture file. Without this, the `resolvedSource` would return undefined/missing.
- I also updated the error message to be less vague. https://github.com/Shopify/eslint-plugin-shopify/issues/122

BPScott (commented):
Nice, working on absolute paths feels like it protects against false-positives when people do odd things with convoluted relative paths (e.g [this](https://github.com/Shopify/eslint-plugin-shopify/issues/119#issuecomment-409695224) should now consistently pass both examples).

Can we add an extra three test cases:

- when a component imports a sibling component using a not the simplest relative path
- when a component's subcomponents import a sibling subcomponent
- when a subcomponent imports its parent's SCSS

The following tests should all pass:

```js
{
  code: `import {someThing} from '../../components/Bar';`,
  parserOptions,
  filename: fixtureFile('basic-app/app/components/Foo/index.js'),
},
{
  code: `import {someThing} from '../Baz';`,
  parserOptions,
  filename: fixtureFile('basic-app/app/components/Foo/components/Bar/index.js'),
},
{
  code: `import {someThing} from '../../Foo.scss';`,
  parserOptions,
  filename: fixtureFile('basic-app/app/components/Foo/components/Bar/index.js'),
}
```

cartogram:
@BPScott updated with the tests you posted.

BPScott (approved):

Issue #73, opened 2018-01-12
Add some more TypeScript linting rules


lemonmade:
- Enums use pascal case
- Don't use .tsx without React

alexandcote:
Most of the issues listed above are now solve. Feel free to reopen an issue if I missed something.

PR #162, opened 2018-09-15
Adds `jsx-require-react` rule

cartogram:
Closes https://github.com/Shopify/eslint-plugin-shopify/issues/73

This rule enforces a React import when writing JSX. Will follow up with updating the readme, configs, and changelog when the general name and approach is approved.

#### Questions

1. Should we also support `const React = require('react')`.
1. I had a previous version of this rule that takes a different approach. Rather than looking at the filename and ImportDeclaration, it gathered variables with the name React. It was far more confusing but interested if anyone thinks that's a better approach here.
1. Always unsure of the name, so let me know if it is clear.

### DIFF
--- a/docs/rules/jsx-require-react.md
+++ b/docs/rules/jsx-require-react.md
@@ -1,0 +1,73 @@
+# Require React import in .jsx and .tsx files. (jsx-require-react)
+
+JSX really just provides syntactic sugar for React's `createElement(component, props, ...children)` function. For example:
+
+```jsx
+<Button color="blue">
+  Click Me
+</Button>
+```
+
+will compile to:
+
+```js
+React.createElement(
+  Button,
+  {color: 'blue'},
+  'Click Me'
+)
+```
+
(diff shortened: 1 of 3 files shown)


docs/rules/jsx-require-react.md
```diff
@@ -0,0 +1,73 @@
...
+React.createElement(
+  'div',
+  {className: 'fancy'},
+  null
+)
+```
+
+## Rule Details
+
+The requirement to import React can easily be missed because it is not directly referenced when authoring JSX. This rule necessitates that React is imported when working in a .tsx or .jsx files.
```
lemonmade, line 39: Can it be easily missed? I thought you always got an error from somewhere about this already :S

kaelig: Agreed, the build or the editor itself would probably tell you there's an issue before even running into a warning from ESLint.

cartogram:
Closing since it doesn't seem like a rule we want to add.

PR #163, opened 2018-09-15
Adds `no-bad-enzyme-features` rule

cartogram:
Part of https://github.com/Shopify/eslint-plugin-shopify/issues/77

#### Questions
1. What are all the [bad parts](https://github.com/Shopify/eslint-plugin-shopify/pull/163/files#diff-82854c898309f636ecdc9af8dded2ae2R71) we want to prevent? 
1. Should this be configurable? probably.
1. better name ideas?

Will follow up with the docs, changelog, readme, etc...

### DIFF
--- a/docs/rules/enzyme/no-bad-features.md
+++ b/docs/rules/enzyme/no-bad-features.md
@@ -1,0 +1,71 @@
+# Disallows a group of Enzyme features that lead to bad practices. (enzyme/no-bad-features)
+
+This rule restricts usuage of the [Enzyme testing utility](https://airbnb.io/enzyme) by preventing a specific set of features that, more often than not, lead to bad testing practices. 
+
+## Rule Details
+
+The disallowed features are grouped into "immutability" and "component-boundaries" as follows:
+
+#### `immutability`
+- `setState`
+- `setContext`
+- `simulate`
+- `simulateError`
+
+#### `component-boundaries`
+- `instance`
+- `state`
+- `ref`
+
+You can learn more about each of these features on the [Enzyme documentation](https://airbnb.io/enzyme).
+
+Examples of **incorrect** code for the `immutability` group of features:
+
+```ts
+const wrapper = mount(<Foo />);
+wrapper.setState({foo: 'bar'});
+wrapper.setContext({bar: 'baz'});
+wrapper.simulate('click');
+wrapper.simulateError(new Error('Explosions!'));
+```
+
+Examples of **incorrect** code for the `component-boundaries` group of features:
+
+```ts
+const wrapper = shallow(<Foo />);
+const instance = wrapper.instance();
+const someState = wrapper.state('someStateKey');
+const someRef = wrapper.ref('someRefAttribute');
+```
+
+### `allow`
+
+```json
+{
+  "enzyme/no-bad-features": [
+    "error",
+    {
+      "allow": ["immutability", "instance"]
+    }
+  ]
+}
+```
+
+This option takes an array of whitelisted features or groups of features that configures the rule to not report their usage. There are nine possible values for this option.
+
+- `"immutability"`
+- `"component-boundaries"`
+- `"setState"`:
+- `"setContext"`
+- `"simulate"`
+- `"simulateError"`
+- `"instance"`
+- `"state"`
+- `"ref"`
+
+
+By default, none of these options are enabled (the equivalent of `{ "allow": [] }`).
+
+## When Not To Use It
+
+If you do not wish to restrict Enzyme features, you can safely disable this rule.
--- a/lib/rules/enzyme/no-bad-features.js
+++ b/lib/rules/enzyme/no-bad-features.js
@@ -1,0 +1,129 @@
+const BAD_FEATURES = {
+  immutability: ['setState', 'setContext', 'simulate', 'simulateError'],
+  'component-boundaries': ['instance', 'state', 'ref'],
+};
+
+module.exports = {
+  meta: {
+    docs: {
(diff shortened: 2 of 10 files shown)


docs/rules/enzyme-no-bad-features.md
```diff
@@ -0,0 +1,36 @@
+# Restricts usage of enzyme to the good parts. (enzyme-no-bad-features)
```
lemonmade: I feel this name is appropriate tbh

lib/rules/enzyme-no-bad-features.js
```diff
@@ -0,0 +1,102 @@
...
+    return {
+      'Program:exit': function() {
+        enzymeWrappers.forEach(checkNodeForInvalidProps);
+      },
+      CallExpression(node) {
+        if (!isEnzymeWrapper(node)) {
+          return;
+        }
+
+        enzymeWrappers.push(node);
```
lemonmade: I don’t see the benefit of pushing to an array and doing it all on exit. What is the difference from just doing it at the AST node itself?

lib/rules/enzyme-no-bad-features.js
```diff
@@ -0,0 +1,102 @@
...
+      },
+    };
+
+    function report(node, prop) {
```
lemonmade: I know this works but IMO air is very confusing to see things after an unguarded return.

lib/rules/enzyme-no-bad-features.js
lemonmade: Never mind I see it now

lib/rules/enzyme-no-bad-features.js
```diff
@@ -0,0 +1,102 @@
...
+
+      const variableDeclaratorParent = findParent(
+        node,
+        (parent) => parent.type === 'VariableDeclarator',
+      );
+
+      if (variableDeclaratorParent) {
+        const references = context
+          .getDeclaredVariables(variableDeclaratorParent)[0]
+          .references.slice(1);
```
lemonmade: Why the slice?

lib/rules/enzyme-no-bad-features.js
```diff
@@ -0,0 +1,102 @@
...
+  return [
+    'setState',
+    'setContext',
+    'setProps',
+    'instance',
+    'state',
+    'simulate',
+    'simulateError',
+    'ref',
+    'context',
```
lemonmade: We still need them to get context, and to set props, in the course of normal tests, so IMO not the right defaults. Agree it should be configurable though.

lemonmade (changes requested):

GoodForOneFare (commented):
I'm oversensitive to computers yelling at me, so messages like "No bad enzyme features" gets my dander up.

Splitting into at least two rules would provide better context for error messages:
- `shopify/enzyme/immutable` (forbids set*, simulate*)
- `shopify/enzyme/component-boundaries` (forbids instance, state, ref, context)

Not keen on the second rule's naming, but you get the idea 😅

cartogram:
@lemonmade this is ready for a review. No rush. Also if you spot what piece of config I'm missing to pass the `all` test, please let me know 🙏

docs/rules/enzyme/no-bad-features.md
```diff
@@ -0,0 +1,71 @@
...
+You can learn more about each of these features on the [Enzyme documentation](https://airbnb.io/enzyme).
+
+Examples of **incorrect** code for the `immutability` group of features:
+
+```ts
+const wrapper = mount(<Foo />);
+wrapper.setState({foo: 'bar'});
+wrapper.setContext({bar: 'baz'});
+wrapper.simulate('click');
+wrapper.simulateError(new Error('Explosions!'));
```
lemonmade, line 29: I don’t know that these two fit as cleanl into immutability. IMO, they are kind of their own class (maybe separate rule?), and the other two here fit in to component boundaries

lib/rules/enzyme/no-bad-features.js
```diff
@@ -0,0 +1,129 @@
...
+  create(context) {
+    const enzymeWrappers = [];
+    const options = context.options[0] || {};
+    const allowed = options.allow || [];
+
+    function report(node, prop) {
+      context.report({
+        node,
+        message:
+          'No bad enzyme features. Avoid using {{prop}} on an enzyme wrapper.',
```
lemonmade: I think we might need or-feature recommendations in place of “don’t do bad”

lib/rules/enzyme/no-bad-features.js
```diff
@@ -0,0 +1,129 @@
...
+        node,
+        (parent) => parent.type === 'VariableDeclarator',
+      );
+
+      if (variableDeclaratorParent) {
+        const references = context
+          .getDeclaredVariables(variableDeclaratorParent)[0]
+          .references.slice(1);
+        references.forEach((reference) =>
+          checkNodeForInvalidProps(reference.identifier),
```
lemonmade, line 80: Seems odd that this is basically recursive?

lemonmade (changes requested):

lib/rules/enzyme/no-bad-features.js
cartogram: How is it odd? Basically want to unwarp the node and check the references for invalid props.

lib/rules/enzyme/no-bad-features.js
cartogram: Yeah good call.

cartogram:
As we move away from enzyme, I don't see a need for this rule anymore.

PR #165, opened 2018-09-19
adding eslint-plugin-graphql as dependency

patsissons:
relates to https://github.com/Shopify/sewing-kit/issues/971 and https://github.com/Shopify/web/pull/6689

This plugin is used by `sewing-kit`. Rather than having sewing kit depend on both this plugin and `eslint-plugin-shopify`, we can instead shuffle `eslint-plugin-graphql` as a dependency of this plugin and pull in all plugins from a single dependency requirement.


lemonmade:
Presumably we'd want to expose a config for GraphQL that just passes a bunch of stuff through to the ESLint plugin rules?

patsissons:
the base graphql config for the plugin (as per the [online docs](https://github.com/apollographql/eslint-plugin-graphql#example-config-when-using-graphqlconfig))
```js
module.exports = {
  parser: 'babel-eslint',
  rules: {
    "graphql/template-strings": ['error', {
      env: 'literal'
    }]
  },
  plugins: [
    'graphql'
  ]
}
```

lemonmade:
Oh right, it doesn't even need schema info. We can probably just have a new config for `graphql` that does this.

patsissons:
on it! just updating the test

patsissons:
@lemonmade should this be added to the `all` config?

lemonmade:
I think so, IIRC `all` is used for detecting missing configuration for rules

patsissons:
how does that work if the rule from the config is not exposed (like the [`jquery`](https://github.com/Shopify/eslint-plugin-shopify/blob/master/lib/config/jquery.js) config)

patsissons:
In order to test the `all` config, we need to add a `graphql@^0.13.0` devDependency (a [requirement of eslint-plugin-graphql](https://github.com/apollographql/eslint-plugin-graphql/blob/master/package.json#L45)) and supply a dummy `.graphqlconfig.yml` and `schema.graphql` file to enable the plugin to full initialize. Unfortunately, the plugin is not smart enough to bail out gracefully if it is included but no graphql configuration exists.

package.json
```diff
@@ -96,6 +97,7 @@
     "eslint-plugin-react": "7.11.1",
     "eslint-plugin-sort-class-members": "1.3.1",
     "eslint-plugin-typescript": "0.12.0",
+    "graphql": "^0.13.0",
```
GoodForOneFare: `eslint-plugin-graphql` declares a `peerDependency` on graphql.  Isn't that enough?

patsissons: i guess we only need it for testing, i can relegate to a dev dep.

GoodForOneFare: Makes sense, ty 👍

GoodForOneFare (approved):
Very much appreciated 👍 Thank you for pulling on this thread, then following through so quickly.

PR #167, opened 2018-09-19
fix getMethodName issues in no-vague-titles rule

patsissons:
this fixes three issues with `getMethodName`
* an `Error: Could not get method name from node.` issue when the linter is parsing an import expression (i.e., `import('./foo')`).
* an `Error: Could not get method name from node.` issue when the linter is parsing a `super` expression (i.e., `super()`).
* a `TypeError: Cannot read property 'name' of undefined` issue when the linter is parsing an objectless call expression (i.e., `foo(bar)()`).

### DIFF
--- a/lib/rules/jest/no-vague-titles.js
+++ b/lib/rules/jest/no-vague-titles.js
@@ -77,13 +77,19 @@ function hasEmptyDescription({arguments: args}) {
 function getMethodName({callee}) {
   switch (callee.type) {
     case 'CallExpression':
-      return callee.callee.object.name;
+      return callee.callee.object
+        ? callee.callee.object.name
+        : callee.callee.name;
     case 'Identifier':
       return callee.name;
     case 'MemberExpression':
       return callee.object.name;
     case 'ArrowFunctionExpression':
       return '';
+    case 'Import':
+      return 'import';
+    case 'Super':
+      return 'super';
     default:
       throw new Error('Could not get method name from node.');
   }
--- a/tests/lib/rules/jest/no-vague-titles.js
+++ b/tests/lib/rules/jest/no-vague-titles.js
@@ -125,6 +125,18 @@ ruleTester.run('no-vague-titles', rule, {
       code: `test.each([['production'], ['staging']])('Includes things for %s clients')`,
       parser,
     },
+    {
+      code: `import('./foo')`,
+      parser,
+    },
+    {
+      code: 'foo(bar)()',
+      parser,
+    },
+    {
+      code: 'class Foo { constructor() { super(); } }',
+      parser,
+    },
   ],
   invalid: [
     {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # Changelog
 
-<!-- ## Unreleased -->
+## Unreleased
+
+### Fixed
+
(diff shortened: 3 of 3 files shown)


lib/rules/jest/no-vague-titles.js
```diff
@@ -77,13 +77,19 @@ function hasEmptyDescription({arguments: args}) {
...
+        ? callee.callee.object.name
+        : callee.callee.name;
     case 'Identifier':
       return callee.name;
     case 'MemberExpression':
       return callee.object.name;
     case 'ArrowFunctionExpression':
       return '';
+    case 'Import':
+      return 'import';
```
GoodForOneFare, line 90: 🤓This is really a keyword, so maybe '' would be more appropriate   Doesn't really matter ¯\\_(ツ)_/¯

tests/lib/rules/jest/no-vague-titles.js
```diff
@@ -125,6 +125,18 @@ ruleTester.run('no-vague-titles', rule, {
...
+    {
+      code: `import('./foo')`,
+      parser,
+    },
+    {
+      code: 'foo(bar)()',
+      parser,
+    },
+    {
+      code: 'super()',
```
GoodForOneFare: Ehhhhh, I'd really like to see this used in a meaningful context.

GoodForOneFare (approved):

lib/rules/jest/no-vague-titles.js
patsissons: yea it doesn't really matter, so i just went with something meaningful if anyone had to debug it.

cartogram (commented):
Thanks for doing this! 🎩 in webgen worked for me.

tests/lib/rules/jest/no-vague-titles.js
patsissons: I'm not sure the test harness supports that, just a simple expression was all i could do. we probably need to limit which files this rule is looking at because the error happens on a file that is not a test.

`src/tools/webpack/dev/client.ts` produces this error:
```ts
export default class Client extends EventEmitter {
  private webpackDevMiddleware: createWebpackMiddleware.WebpackDevMiddleware | null;
  private listenerManager: ListenerManager | null;

  constructor(
    private config: webpack.Configuration,
    private workspace: Workspace,
    private dashboard: Dashboard,
    private runner: Runner,
    private options: Partial<Options> = {},
  ) {
    super(); // <-----------------------------------------------
    this.webpackDevMiddleware = null;
    this.listenerManager = null;
  }
  //...
}
```

patsissons: @GoodForOneFare do you have any suggestions for this one? or should we leave it as is and deal with the input file filter later?

lib/rules/jest/no-vague-titles.js
patsissons: for added context, this specific case is when it is parsing an import as a function, i.e. `const foo = await import('./foo');` so the function name `import` is somewhat accurate.

tests/lib/rules/jest/no-vague-titles.js
GoodForOneFare: This should work:
```
    {
      code: 'class Foo { constructor() { super(); } }',
      parser: 'babel-eslint',
    },
```

patsissons: I swear i tried something like that. clearly i was imagining it 😁

PR #170, opened 2018-09-21
updating prettier config to enable the prettier/prettier rule

### LINKED ISSUE #169: prettier config does not enable prettier linting
#121 removed all inline prettier config to prevent overriding local `.prettierrc` config. We still need to define `'prettier/prettier': 'error'` in the `prettier` config to enable linting with `prettier`.

patsissons:
When the `plugin:shopify/prettier` config is included in the eslint configuration, prettier linting is not active unless the configuration also include `'prettier/prettier': 'error'`. This PR adds the bare minimum config to enable prettier linting without overriding local file configuration.

resolves #169 
relates to #118 and #121 

### 🎩 Notes

There are new tests (in `tests/lib/config/prettier.js` and `tests/lib/config/typescript-prettier.js`) that confirm we can extend `plugin:shopify/prettier` and `plugin:shopify/typescript-prettier` respectively and validate that prettier rules are enforced on linted source files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,16 @@
 # Changelog
 
-<!-- ## Unreleased -->
+## Unreleased
+
+### Fixed
+
+* Updated `plugin:shopify/prettier` to enable prettier linting. ([#170](https://github.com/Shopify/eslint-plugin-shopify/pull/170))
 
 ## [24.2.0] - 2018-09-21
 
 ### Added
 
-* Added `plugin:shopify/graphql` to module index. ([#000](https://github.com/Shopify/eslint-plugin-shopify/pull/000))
+* Added `plugin:shopify/graphql` to module index. ([#168](https://github.com/Shopify/eslint-plugin-shopify/pull/168))
 
 ### Fixed
 
--- a/lib/config/prettier.js
+++ b/lib/config/prettier.js
@@ -1,9 +1,11 @@
+const merge = require('merge');
+
 module.exports = {
   extends: ['prettier'],
 
   plugins: ['prettier'],
 
-  rules: {
+  rules: merge(require('./rules/prettier'), {
     // rules to disable to prefer prettier
     'shopify/class-property-semi': 'off',
     'shopify/binary-assignment-parens': 'off',
@@ -21,5 +23,5 @@ module.exports = {
     // Special rule for 'no-unexpected-multiline'
     // https://github.com/prettier/eslint-config-prettier/blob/5399175c37466747aae9d407021dffec2c169c8b/README.md#no-unexpected-multiline
     'no-unexpected-multiline': 'error',
-  },
+  }),
 };
--- a/lib/config/typescript-prettier.js
+++ b/lib/config/typescript-prettier.js
@@ -4,12 +4,5 @@ module.exports = {
   rules: {
     // rules to disable to prefer prettier
     'typescript/member-delimiter-style': 'off',
-
-    'prettier/prettier': [
-      'error',
-      {
-        parser: 'typescript',
-      },
-    ],
   },
 };
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
     "babel-cli": "^6.26.0",
     "babel-core": "^6.26.3",
     "babel-preset-shopify": "^16.5.0",
+    "chai": "^4.0.0",
     "eslint": "^5.4.0",
     "eslint-index": "^1.4.0",
     "eslint-plugin-shopify": "file:./.",
--- a/tests/fixtures/typescript-prettier/package.json
+++ b/tests/fixtures/typescript-prettier/package.json
@@ -1,0 +1,5 @@
+{
+  "eslintConfig": {
+    "extends": "plugin:shopify/typescript-prettier"
+  }
+}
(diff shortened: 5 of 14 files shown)


CHANGELOG.md
```diff
@@ -1,12 +1,16 @@
...
+### Fixed
+
+* Updated `plugin:shopify/prettier` to enable prettier linting. ([#170](https://github.com/Shopify/eslint-plugin-shopify/pull/170))
 
 ## [24.2.0] - 2018-09-21
 
 ### Added
 
-* Added `plugin:shopify/graphql` to module index. ([#000](https://github.com/Shopify/eslint-plugin-shopify/pull/000))
+* Added `plugin:shopify/graphql` to module index. ([#168](https://github.com/Shopify/eslint-plugin-shopify/pull/168))
```
patsissons, line 13: don't mind this, just fixing missing markup.

lib/config/typescript-prettier.js
```diff
@@ -1,15 +1,10 @@
+const merge = require('merge');
+
 module.exports = {
   extends: ['plugin:shopify/prettier'],
 
-  rules: {
+  rules: merge(require('./rules/typescript-prettier'), {
```
patsissons: I refactored the typescript variant at the same time to import the config. maybe not necessary, but should be no harm.

lib/config/prettier.js
```diff
@@ -1,9 +1,11 @@
+const merge = require('merge');
+
 module.exports = {
   extends: ['prettier'],
 
   plugins: ['prettier'],
 
-  rules: {
+  rules: merge(require('./rules/prettier'), {
```
BPScott, line 8: I think it's a lot cleaner to avoid this indirection and just have `'prettier/prettier': ["error"]` defined as a rule here.

As we don't re-use that `rules/prettier` config file elsewhere I don't think it's worth abstracting out given that we never re-use it.

BPScott (dismissed):

lib/config/prettier.js
patsissons: the rule is referenced in the `all.js` config (for testing), that's the only reason for splitting it out afaik. the `all` config was already pulling in the `prettier` plugin, but just didn't have any related rule to enable the plugin.

BPScott: Ah gotcha. This is consistent with how all the other configs get defined. +1 for consistency.

GoodForOneFare (approved):
Good find 👍 Mystery solved 😆

package.json
```diff
@@ -20,7 +20,7 @@
     "check": "yarn run lint && yarn test",
     "rules-status": "eslint-index lib/config/all.js --format table",
     "rules-omitted": "eslint-index lib/config/all.js --status omitted",
-    "lint": "eslint . --max-warnings 0 --format codeframe",
+    "lint": "eslint . --max-warnings 0 --format codeframe --ignore-pattern tests/fixtures",
```
patsissons: I had to change this to prevent the repo lint from complaining about intentionally bad files in fixtures that I need to still result in errors when we lint from a test.

patsissons:
@BPScott and @GoodForOneFare give this another glance to make sure we're good on this change (now that it has gone through some adjustments).

package.json
BPScott: Can you move this into the .eslintignore file: https://github.com/Shopify/eslint-plugin-shopify/blob/master/.eslintignore

lib/config/typescript-prettier.js
```diff
@@ -4,12 +4,5 @@ module.exports = {
   rules: {
     // rules to disable to prefer prettier
     'typescript/member-delimiter-style': 'off',
```
BPScott, line 6: It looks like prettier changed to use ; as member delimiters, to match the eslint rule way back in v 1.4.1 as per https://github.com/prettier/prettier/pull/1918, so I think we can remove this too. But that can be done in a different PR.

BPScott (approved):
One small nit about where to put ignore files but other than that, looks good.

lib/config/typescript-prettier.js
patsissons: makes sense. let's shuffle that in a different PR to keep this one focused.

package.json
patsissons: I'm not sure if I can use the global ignores because it will affect the tests that are validating the files we want CI `lint` to ignore but not CI `test`

tests/fixtures/typescript-prettier/package.json
```diff
@@ -0,0 +1 @@
+{}
```
GoodForOneFare: I'd prefer the setup in `.eslintrc.js` to be moved into an `eslintConfig` in here.  One less file ¯\\_(ツ)_/¯

tests/lib/config/prettier.js
```diff
@@ -0,0 +1,21 @@
+/* eslint-env node, mocha */
+/* eslint-disable flowtype/require-valid-file-annotation */
+import {expect} from 'chai';
+import {execSync} from 'child_process';
+
+// eslint-disable-next-line shopify/strict-component-boundaries
+import {fixtureFile} from '../../utilities';
+
+describe('config', () => {
+  describe('typescript-prettier', () => {
```
GoodForOneFare: `typescript-prettier` seems off here.

GoodForOneFare (approved):
Couple of minor comments, but 👍 :shipit: overall.

package.json
patsissons: new plan, `--no-ignore` can override the global ignores for these tests

PR #171, opened 2018-09-26
restoring prettier parser override in the typescript-prettier config

patsissons:
Fixes issue with `prettier` plugin using the incorrect parser when the `graphql` plugin is also enabled.

### DIFF
--- a/lib/config/typescript-prettier.js
+++ b/lib/config/typescript-prettier.js
@@ -4,5 +4,16 @@ module.exports = {
   rules: {
     // rules to disable to prefer prettier
     'typescript/member-delimiter-style': 'off',
+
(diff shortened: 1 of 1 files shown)


BPScott (approved):
This restores code that was deleted in https://github.com/Shopify/eslint-plugin-shopify/pull/170/files#diff-56a2d129bcbecca23453c869c1300973L7

This revert is the easiest path to reclaim old behaviour. 

@patsissons  and I have been pairing on this today and we have a better, more long-term fix for this based upon [investigation](https://github.com/prettier/eslint-plugin-prettier/issues/81#issuecomment-424550756). Pat shall put that up, with a bunch of regression tests once this is merged.

GoodForOneFare (approved):

PR #173, opened 2018-09-27
updating prettier and react configs

patsissons:
This PR deprecates `plugin:shopify/typescript-prettier` and `plugin:shopify/typescript-react` in favour of using overrides in `plugin:shopify/prettier` and `plugin:shopify/react` respectively. Additionally, `plugin:shopify/typescript` focuses any applied rules on typescript files only.

`prettier` (and related tooling) is also getting an upgrade.


patsissons:
@BPScott have a glance at this when you have a moment and make sure it's aligned with what we worked on the other day. The only notable change I made was to just completely disable prettier for graphql files (since the `babylon` parser doesn't really do any formatting of the `eslint-plugin-graphql`-_ified_ content).

lib/config/prettier.js
```diff
@@ -24,4 +24,21 @@ module.exports = {
...
+  overrides: [
+    {
+      // force babylon parser for graphql files
+      files: ['*.graphql', '*.gql'],
+      rules: {
+        'prettier/prettier': 'off',
+      },
+    },
+    {
+      // force typescript parser for TypeScript files
```
BPScott: This shouldn't be needed as eslint-plugin-prettier 2.7.0 (and really anything after 2.4.0) shall infer a parser based upon the the file extension - including using the typescript parser for ts/tsx files

lib/config/prettier.js
```diff
@@ -24,4 +24,21 @@ module.exports = {
     // https://github.com/prettier/eslint-config-prettier/blob/5399175c37466747aae9d407021dffec2c169c8b/README.md#no-unexpected-multiline
     'no-unexpected-multiline': 'error',
   }),
+
+  overrides: [
+    {
+      // force babylon parser for graphql files
```
BPScott: This comment is doesn't reflect the code.

eslint-plugin-prettier v2.7.0 forces the babylon parser for graphql files internally (big kudos to their maintainer for the rapid release of my PR), so we no longer need to explicitly turn that off to stop things from crashing.

You're right that we would run prettier over these files and no errors would be reported thanks to eslint-plugin-graphql [suppressing output for rules other than its own](https://github.com/apollographql/eslint-plugin-graphql/blob/fe6f6d7627605d2b1259bf19cba6985bf53fd671/src/index.js#L515-L518) (per our convo yesterday). I'm torn between having extra config to save a slight bit of time and having no config to be ready for a future where eslint-plugin-graphql exposes other rule violations and supports autofixing.

lib/config/typescript-prettier.js
```diff
@@ -1,19 +1,4 @@
 module.exports = {
+  // DEPRECATED: plugin:shopify/prettier is sufficient now
```
BPScott: `DEPRECATED: Please use plugin:shopify/prettier.  plugin:shopify/typescript-prettier shall be removed in a future major release` to  lay out a future plan perhaps?

lib/config/typescript-prettier.js
```diff
@@ -1,19 +1,4 @@
...
   extends: ['plugin:shopify/prettier'],
-
-  rules: {
-    // rules to disable to prefer prettier
-    'typescript/member-delimiter-style': 'off',
```
BPScott, line 6: Disabling this rule should still be present (but in an overrides section prettier) as we want to defer to prettier rather than risk configuration of conflicting with prettier's output. (Note the default value when enabling this rule matches prettier, but people may have changed it to something else)

lib/config/typescript-react.js
```diff
@@ -1,11 +1,4 @@
...
-
 module.exports = {
-  extends: [
-    // react should come before typescript
-    'plugin:shopify/react',
-    'plugin:shopify/typescript',
-  ],
-
-  rules: merge(require('./rules/typescript-react')),
+  // DEPRECATED: plugin:shopify/react is sufficient now
```
BPScott: `DEPRECATED: Please use plugin:shopify/react. plugin:shopify/typescript-react shall be removed in a future major release` to lay out a future plan perhaps?

package.json
```diff
@@ -49,11 +49,10 @@
       "prettier/prettier": [
         "error",
         {
-          "singleQuote": true,
```
BPScott, line 52: I believe all these settings can be removed, as they should be inferred from `.prettierrc` if not explicitly overridden.

BPScott (changes requested):
Nice! I'm loving these test cases.

In addition to the inline comments we should also update the README to stop mentioning typescript-prettier / typescript-react, and update the new mechanism enabling for typescript rules

lib/config/prettier.js
patsissons: my vote is to disable for now and re-enable when support is added. That way things don't magically break when a plugin unrelated to this config is updated.

lib/config/typescript-prettier.js
patsissons: ~would leaving this rule in our config not override what is defined in a `.prettierrc` file?~ gah, withdrawn. tired eyes did not notice this is a lint rule, not a prettier rule.

BPScott: The typescript ruleset [configures "typescript/member-delimiter-style"](https://github.com/Shopify/eslint-plugin-shopify/blob/4fe4585365b1a91f9bfedfd3beccf15ad68a526d/lib/config/rules/typescript.js#L87). When using prettier, we want to disable that eslint rule, and instead trust whatever prettier outputs.

BPScott:
What do you think to updating https://github.com/Shopify/eslint-plugin-shopify/blob/master/lib/config/typescript.js so that it only applies the typescript rules to *.ts and *.tsx files? (To match the behaviour of typescript-react rules only being applied to *.ts/*.tsx files)

As it currently stands force all js files to use the typescript parser and rules, which means [we sometimes have to disable those rules in js files](https://github.com/Shopify/polaris-react/pull/2240/files#diff-1dc6ee56b778cd91e0327b52aaeaa1b9R53)

patsissons:
seems reasonable to me. Should probably do the same for the graphql config, but i'll do that in a different PR. I'll do the TS overrides now though.

patsissons:
`plugin:shopify/prettier`, `plugin:shopify/react`, and `plugin:shopify/typescript` now all use overrides to focus typescript rules on only typescript files.

GoodForOneFare:
Subtle but huge quality of life changes in here.  I love this PR ❤️

BPScott (approved):
Beautiful!

eslint-plugin-prettier 3.0 was released earlier, it contains mostly dropping support for EOL node and eslint versions and pragmas (which we don't use) so should be a drop-in replacement for us. It'd be good to upgrade to that.

Then I will shut up and this can be merged :D

patsissons:
Should we also upgrade `eslint-config-prettier`? or defer that to another PR/release.

BPScott:
it doesn't look like there's anything In the newer versions of eslint-config-prettier that I care about right now, so that can be done later IMO.

patsissons:
sounds good to me, shipping shortly...

PR #174, opened 2018-09-30
Adds `no-export-all` rule

cartogram:
First of a few rules that I will add that deal with consistent exporting.

Looking for general approval here (name, error message, etc...). Will follow up with the docs, configs, etc... once I receive confirmation this is a rule we actually want.

Things to review so far.
- Name (`no-export-all`)
- [Approach](https://github.com/Shopify/eslint-plugin-shopify/pull/174/files#diff-70e4aed6d9965b52c7315c1f79c1d253R16)
- [Tests](https://github.com/Shopify/eslint-plugin-shopify/pull/174/files#diff-c7f5fb7a1786a4286d89460b03b725d9R48)

@GoodForOneFare why did the `yarn.lock` appear in the diff?

### Update

Added everything ^, this is ready for a real review now.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1909,7 +1909,7 @@ eslint-plugin-react@7.11.1:
     prop-types "^15.6.2"
 
 "eslint-plugin-shopify@file:./.":
-  version "25.1.0"
+  version "26.0.0"
   dependencies:
     babel-eslint "9.0.0"
     eslint-config-prettier "3.0.1"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Added
 * `shopify/no-ancestor-directory-import` ([#149](https://github.com/Shopify/eslint-plugin-shopify/pull/149))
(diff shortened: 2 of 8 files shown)


lemonmade (commented):
I am on board with this rule, I think we just need the message to be more helpful/ explain *why* this is an antipattern.

marutypes:
Personally I’d rather spend as little time updating a million `index.ts` files as possible and I don’t think this pattern ever actually causes problems for anyone, so I think this feels a little puritanical to me. I do get the appeal, and I don't feel super strongly, but it feels mildly :/

yarn.lock
```diff
@@ -1909,7 +1909,7 @@ eslint-plugin-react@7.11.1:
     prop-types "^15.6.2"
 
 "eslint-plugin-shopify@file:./.":
-  version "25.1.0"
+  version "26.0.0"
```
ismail-syed, line 1912: This repo lints itself using itself. When we release a new version, this file gets updated with the new version of itself. We probably forgot to do this in the last release.

marutypes: 🔁

PR #175, opened 2018-10-01
adds `consistent-export-style` rule

cartogram:
Closes part of: https://github.com/Shopify/web-foundation/issues/130

This is a first pass at enforcing a consistent export style in `components/index` files. At this point, I am looking for feedback on the general approach as this rule is not entirely straightforward.

According to this rule, the following code would be invalid because there are multiple exports from the same source.

```ts
export {default as BulkDiscountCode} from './BulkDiscountCode';
export {Props as BulkDiscountCodeProps} from './BulkDiscountCode';
export {DiscountCodesAppDescriptor} from './BulkDiscountCode';
```
The following would be invalid because shape of the specifiers does not match our ideal shape:
```ts
export {
  default as Button,
  Props as ButtonProps,
} from './NotAButton';
```

The ideal shape would be as follows:

```ts
export {
  default as Button,
  Props as ButtonProps,
  SomeRandomThing, // this part doesn't matter
} from './Button';
```

**Question**
* Should we add logic to enforce a strict syntax for exports in `compoentName/index`, or should that be a separate rule. It would be pretty easy to combine them here, but then we might want to provide a config for enabling one or the other individually.

### DIFF
--- a/lib/rules/consistent-export-style.js
+++ b/lib/rules/consistent-export-style.js
@@ -1,0 +1,106 @@
+import {sep, basename, extname} from 'path';
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Enforces consistency in how components are exported',
+      category: 'Best Practises',
+      recommended: true,
+    },
+    schema: [],
+  },
+
+  create(context) {
+    const componentExportDecalations = [];
+
+    function isComponentsIndexFile() {
+      const pathParts = context.getFilename().split(sep);
+      const isIndexFile =
+        basename(
+          pathParts[pathParts.length - 1],
+          extname(pathParts[pathParts.length - 1]),
+        ) === 'index';
+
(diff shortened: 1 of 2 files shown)


lemonmade:
So I think there are probably different rules that are needed here:

1. Don't bother checking for duplicate exports from the same place, we have a rule for that already.
2. We probably want a rule that if the only thing that happens with an import is that it gets exported, you just do the export statement. So, `import Button from './Button'; export default Button;` => `export {default} from './Button';`, and similar for named exports
3. With respect to component files specifically, I think we might not need to do anything, other than to ensure that if `Props` is re-exported from a components/index file, it has a name matching `<ComponentName>Props` (this might not be necessary either as I feel we should probably stop exporting the `Props` type anyways)

cartogram:
Closing this for now. Will revisit at a later point.

PR #178, opened 2018-10-02
Include TSAsExpression in getMethodName

cartogram:
Fixes an issue in web. More details found in this PR: https://github.com/Shopify/web/pull/7013

### DIFF
--- a/lib/rules/jest/no-vague-titles.js
+++ b/lib/rules/jest/no-vague-titles.js
@@ -84,14 +84,8 @@ function getMethodName({callee}) {
       return callee.name;
     case 'MemberExpression':
       return callee.object.name;
-    case 'ArrowFunctionExpression':
-      return '';
-    case 'Import':
-      return 'import';
-    case 'Super':
-      return 'super';
     default:
-      throw new Error('Could not get method name from node.');
+      return false;
   }
 }
 
--- a/tests/lib/rules/jest/no-vague-titles.js
+++ b/tests/lib/rules/jest/no-vague-titles.js
@@ -4,7 +4,9 @@ const rule = require('../../../../lib/rules/jest/no-vague-titles');
 const ruleTester = new RuleTester();
 
 require('babel-eslint');
+require('typescript-eslint-parser');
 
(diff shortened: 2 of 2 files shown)


lib/rules/jest/no-vague-titles.js
```diff
@@ -90,6 +90,8 @@ function getMethodName({callee}) {
       return 'import';
     case 'Super':
       return 'super';
+    case 'TSAsExpression':
+      return '';
```
patsissons: I'm inclined to get rid of this, `Super`, and `Import` and refactor into two functions so that callers expecting a valid method name will experience a thrown error and callers who don't care just receive null. might also be handy to emit the `callee.type` inside the error message.

cartogram: @patsissons Yeah, I'm in agreement there. Will update this PR to do that.

lib/rules/jest/no-vague-titles.js
```diff
@@ -84,14 +84,8 @@ function getMethodName({callee}) {
...
       return callee.object.name;
-    case 'ArrowFunctionExpression':
-      return '';
-    case 'Import':
-      return 'import';
-    case 'Super':
-      return 'super';
     default:
-      throw new Error('Could not get method name from node.');
+      return '';
```
cartogram: I don't know what benefit we get by throwing here. We can just check for the things we care about and return no method name for the things we don't.

patsissons: I'm guessing only once we branch into `containsVagueWord` do we absolutely need the method name.

lemonmade: I think maybe return false instead here so it is significantly different than the other return values

lemonmade (approved):

lib/rules/jest/no-vague-titles.js
marutypes: alternatively: return a symbol named something like `NO_METHOD_NAME`, then instead of falsiness just check against that.

marutypes (approved):
👍

PR #187, opened 2018-10-23
Enable allowWholeFile for eslint-comments/disable-enable-pair

### LINKED ISSUE #183: [rfc] allowWholeFile in eslint-comments/disable-enable-pair
#164 added eslint-comments,

I think it might be useful to set `"eslint-comments/disable-enable-pair": ["error", {"allowWholeFile": true}]` to allow for disabling a rule for the whole file. It's often useful in scripts to disable the no-console rule across a whole file for instance

See https://github.com/mysticatea/eslint-plugin-eslint-comments/blob/master/docs/rules/disable-enable-pair.md#options

BPScott:
This allows an eslint-disable directive without a matching close when it
appears before all code in the file. This allows you to disable rules
for a whole file, such as disabling no-console in scripts.

Fixes #183

### DIFF
--- a/lib/config/rules/eslint-comments.js
+++ b/lib/config/rules/eslint-comments.js
@@ -2,7 +2,7 @@
 
 module.exports = {
   // Require a eslint-enable comment for every eslint-disable comment
-  'eslint-comments/disable-enable-pair': 'error',
+  'eslint-comments/disable-enable-pair': ['error', {allowWholeFile: true}],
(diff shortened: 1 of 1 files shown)


cartogram (approved):
I agree with this. 👏

cartogram:
No changelog required since there was never a release that included this plugin.

GoodForOneFare (approved):

PR #188, opened 2018-10-23
Except arrow functions from declaration rule

sambostock:
Currently, this is frowned upon by the ESLint

```typescript
const MyComponent: React.SFC<Props> = ({foo}) => (<div>{foo}</div>);
```

However, this syntax offers nice semantics in terms of typing `Props`, as well as ensuring the return type is as expected.

The alternative, which ESLint expects (since we aren't using `this`, and there is therefore "no reason" to use an Arrow Function) is

```typescript
function MyComponent({foo}: Props) {
  return (<div>{foo}</div>);
}
```

With this notation,
- the `Props` type must be next to the parameter destructuring, potentially crowding things
- props from `IntrinsicAttributes` must be explicitly added to the `Props` interface
- a separate type must be used for the return value (if we want to validate the return type, which `React.SFC` gives us for free)
	- it is also not immediately obvious what that type should be (`React.Element`? `ReactNode`? something else?)

**This change whitelists arrow functions, such that they can be used as described above.**

### DIFF
--- a/lib/config/rules/stylistic-issues.js
+++ b/lib/config/rules/stylistic-issues.js
@@ -32,7 +32,7 @@ module.exports = {
   // Don't require function expressions to have a name
   'func-names': 'off',
   // Enforces use of function declarations or expressions
-  'func-style': ['error', 'declaration'],
+  'func-style': ['error', 'declaration', {allowArrowFunctions: true}],
   // enforce consistent line breaks inside function parentheses
   'function-paren-newline': ['error', 'consistent'],
   // Blacklist certain identifiers to prevent them being used


sambostock:
@lemonmade Not sure if this should have a test somewhere to validate it is the correct setting?

adamcarchibald (approved):

lemonmade (approved):
Can you update https://github.com/Shopify/javascript#10.2 to include this loosening of the restriction?

sambostock:
Good call, @lemonmade. Started updating it in https://github.com/Shopify/javascript/pull/270. Got some questions, but they can be discussed there.

Good to merge this?

GoodForOneFare (approved):
🚢

kaelig:
Can you add this change under the appropriate changelog version?

PR #189, opened 2018-10-25
Use a stricter import order

BPScott:
A while ago we talked about [ordering imports in #web-foundation-tech](https://shopify.slack.com/archives/CCNRSKLTH/p1538573438000100).

General opinion was alphabetic ordering is annoying but grouping types of imports - node buildins, vs npm modules vs relative paths etc - would be handy as we already aim to do that anyway. 

This updates our import/order rule to be a bit stricter around forcing ordering based upon the type of import to the order goes internal imports (`components/Blah`) then parents (`../Blah`) then siblings (`./Blah`) instead of allowing those to be intermingled.

No ordering inside those groups is enforced, which shall keep the "no alphabets" group happy. And yes this has an autofix.

 See https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md for more info

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 ### Changed
 
 * Updated `plugin:shopify/prettier`, `plugin:shopify/react`, and `plugin:shopify/typescript` to use `overrides` ([#173](https://github.com/Shopify/eslint-plugin-shopify/pull/173))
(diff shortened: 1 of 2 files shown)


cartogram (approved):
The configuration for this rule is interesting. Changes LGTM 👍. Give the people what they want.

lemonmade (approved):

PR #195, opened 2018-10-26
downgrading to eslint-plugin-graphql 2.1.0-0

patsissons:
The current version of `eslint-plugin-graphql` (`2.1.1`) [does not properly support multiple schemas](https://github.com/apollographql/eslint-plugin-graphql/pull/178) due to the schema caching mechanism. This is a rollback to `2.1.0-0` before the caching was introduced. This can be upgraded again once [caching is updated](https://github.com/apollographql/eslint-plugin-graphql/pull/179) to support multiple schemas.

blocked on #196


GoodForOneFare (approved):

BPScott:
Annoyingly it looks like the caching was introduced [between v2.1.0 and v2.1.0-0](https://github.com/apollographql/eslint-plugin-graphql/compare/v2.1.0-0...v2.1.0), so we've got to do the slightly funky version number

BPScott (approved):

PR #197, opened 2018-10-27
Set up import parsers so import/* rules work in typescript

BPScott:
It turns out you need some extra config in order for some of the import
checking rules to do their thing with TypeScript.

To tophat:

* Check out a project that already uses eslint-plugin-shopify
* `yarn add --dev eslint-import-resolver-typescript`
* Add the settings to your eslint config.
* Run eslint. Marvel at all those new errors.

### DIFF
--- a/lib/config/typescript.js
+++ b/lib/config/typescript.js
@@ -12,6 +12,15 @@ module.exports = {
     sourceType: 'module',
   },
 
+  settings: {
+    'import/parsers': {
+      'typescript-eslint-parser': ['.ts', '.tsx'],
+    },
+    'import/resolver': {
+      typescript: {},
+    },
+  },
+
   overrides: [
     {
       files: ['*.ts', '*.tsx'],
--- a/package.json
+++ b/package.json
@@ -70,6 +70,7 @@
   "dependencies": {
     "babel-eslint": "9.0.0",
     "eslint-config-prettier": "3.0.1",
+    "eslint-import-resolver-typescript": "^1.1.1",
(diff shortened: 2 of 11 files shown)


lemonmade (approved):
This is great, good find. I think we still need some more configuration for our default project setups (resolving `packages/*` and so forth), but this is a great start.

BPScott:
>  I think we still need some more configuration for our default project setups (resolving packages/* and so forth)

Adding `"typescript": {},` shall automatically look into the top-level tsconfig.json and read the paths and baseUrl from there so I think this works out the box.

I just tried adding `import Asset from '@shopify/sewing-kit-server'` into /packages/sewing-kit-server/assets in web and it gives me a "dependency cycle detected" warning.

lemonmade:
Oh, awesome. A great start and finish then :P

patsissons (approved):
nice work!

patsissons:
will this be a point release? should i try and roll it into https://github.com/Shopify/sewing-kit/pull/1038 ?

BPScott:
This feels like a minor release (26.1.0) - technically it's fixing a config that /should/ have always been working, but it's going to introduce a lot of new linting messages.

Rolling it into the new SK release sounds like a good plan.

BPScott:
Oh it's also worth pointing out that [this also currently doesn't work for JS files when using the babel-eslint parser due to a bug upstream](https://github.com/benmosher/eslint-plugin-import/issues/1166). but [there's an open PR to fix that](https://github.com/benmosher/eslint-plugin-import/pull/1218)

PR #203, opened 2018-11-02
no-vague-titles no longer flags when call is used

brendo:
The `jest/no-vague-titles` rule flagged the following test name as invalid:

`closing the modal does not call empty when emptyOnClose is false`

I've adjusted the regex to cater for when `all` is part of a word 👍

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 <!-- ## Unreleased -->
 
+### Fixed
+
+* `jest/no-vague-titles` no longer flags when the word `call` is used.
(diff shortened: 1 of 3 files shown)


BPScott:
Would word boundaries be slightly simpler? https://www.regular-expressions.info/wordboundaries.html

brendo:
Good suggestion! Updated 😄

cartogram (approved):
💯Nice!

BPScott (approved):

brendo:
Thanks 👍

Issue #181, opened 2018-10-13
Add more rules that enforce testing best practises


cartogram:
I think most things here can be turned into a rule. https://github.com/Shopify/web-foundation/blob/master/Best%20practices/Jest.md#best-practices

- [ ] mock-order
- [ ] mock-full-module
- [ ] no-bad-mock-features

PR #204, opened 2018-11-02
adds `no-all-mocks-methods` rule

jaxee:
Closes part of https://github.com/Shopify/eslint-plugin-shopify/issues/181
For more information see: https://github.com/Shopify/web-foundation/blob/master/Best%20practices/Jest.md#best-practices

Do not use `jest.resetAllMocks()`, `jest.clearAllMocks()`, `jest.restoreAllMocks()`, and `jest.resetModules()`.

> Why? These methods are overly broad and do not have explicit connections to the contents of your test file. Instead, reset/ clear/ restore individual mocks that are set up explicitly for the purposes of your test suite.

First time adding a rule, feedback welcomed 🙏 

cc @djirdehh

### DIFF
--- a/docs/rules/jest/no-all-mocks-methods.md
+++ b/docs/rules/jest/no-all-mocks-methods.md
@@ -1,0 +1,35 @@
+# Disallows jest allMocks methods.
+
+This rule discourages the use of overly broad Jest methods such as `resetAllMocks`, `clearAllMocks`, `restoreAllMocks` and `resetModules`.
+
+## Rule Details
+
+These methods are discouraged because there should be explicit connections to the contents of your test file. Mocks should be reset/cleared/restored individually based on the purpose of your test suite.
+
+The following patterns are considered warnings:
+
+```js
+jest.resetAllMocks();
+```
+
+```js
+jest.clearAllMocks();
+```
+
+```js
+jest.restoreAllMocks();
+```
+
+```js
+jest.resetModules();
+```
+
+The following patterns are not warnings:
+
+```js
+jest.mock();
+```
+
+## Further Reading
+
+- [Shopify Jest Best Practices](https://github.com/Shopify/web-foundation/blob/master/Best%20practices/Jest.md#best-practices)
--- a/lib/rules/jest/no-all-mocks-methods.js
+++ b/lib/rules/jest/no-all-mocks-methods.js
@@ -1,0 +1,35 @@
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Disallows jest allMocks methods.',
+      category: 'Best Practices',
+      recommended: false,
+      uri:
+        'https://github.com/Shopify/eslint-plugin-shopify/blob/master/docs/rules/jest/no-all-mocks-methods.md',
+    },
+  },
+
+  create(context) {
+    return {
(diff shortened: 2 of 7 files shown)


docs/rules/jest/no-all-mocks-methods.md
```diff
@@ -0,0 +1,35 @@
...
+```
+
+```js
+jest.resetModules();
+```
+
+The following patterns are not warnings:
+
+```js
+jest.mock();
```
jaxee, line 30: Is there a better way to describe here the best practices?

cartogram: I think this is fine, you could go one step further and fill in the alternative for each method you are restricting here. For example, showing a full example using `mockReset();` as the alternative to `resetAllMocks`.

docs/rules/jest/no-all-mocks-methods.md
```diff
@@ -0,0 +1,35 @@
...
+```
+
```
cartogram, line 32: Usually, there is a note here about when to its safe to disable this rule. Its pretty template-y but I would add it for consistency.

lib/rules/jest/no-all-mocks-methods.js
```diff
@@ -0,0 +1,35 @@
...
+  },
+
+  create(context) {
+    return {
+      Identifier(node) {
+        if (isInvalidMocks(node.name)) {
+          context.report({
+            node,
+            message:
+              'Do not use {{method}} or related methods that are not explicit to your test. Instead, target individual mocks.',
```
cartogram: I think you could be even more specific here. Suggestion:`Do not use {{method}} or related methods that are not explicit to a single mock. Instead, clear, reset and restore mocks individually.`

cartogram (approved):
Sorry it took me so long to review this. 😞 

Nothing major on my end, this looks great! It was very cool working with you on this one.

tests/lib/rules/jest/no-all-mocks-methods.js
```diff
@@ -0,0 +1,41 @@
...
+  ];
+}
+
+ruleTester.run('no-all-mocks-methods', rule, {
+  valid: [
+    {
+      code: `jest.mock()`,
+    },
+    {
+      code: `jest.fn()`,
```
cartogram, line 13: This rule is fine as is, but you could check that the method is chained off the `jest` object. Right now, if there were a method in your application called `clearAllMocks` (completely unrelated to jest) that would be considered invalid. This is highly unlikely, but you could alter this rule to prevent against it.

At the same time, if you solve that problem, a second problem could occur if the methods were destructured off the jest object (`const {clearAllMocks} = jest;` or `const clearAllMocks = jest.clearAllMocks` or even `somethingTotallyDifferent = jest.clearAllMocks`). I think your rule will catch this right now.

You could protect against all of the above at the same time, but it is a lot more complex and I don't anticipate too many problems with the rule as is. However, it would be a fun challenge. I am also more than happy to pair again on it.

cartogram:
@jaxee Thanks for your help with this! 🙏 I am going to quickly wrap up the final things so that we can get this out of the PR list. 🙂

PR #219, opened 2019-02-09
Add images-no-direct-imports rule

BPScott:
The Polaris team has been doing some research into how icons (and kinda images in general) get used.

One thing that was identified was that we strongly tend towards putting images in a dedicated folder within a component (overwhelmingly either `images`, `icons` or `illustrations`) then create an index file for that folder. Then within the component you import from that index file.

Once https://github.com/Shopify/web/pull/10568 is merged then the stats will be: 494 svg files are imported into an index, 36 are imported elsewhere.

`images-no-direct-imports` is a new rule to enforce the convention of creating an index file by disallowing imports of image files into any file that is not the index file in the same directory as the svg resides.

---

Open questions:

- What's the preferred convention for namespaced rules? A prefix to the filename (e.g. react-type-state) or create a folder (e.g. typescript/prefer-singular-enums)
- Whats the preferred naming between 'no-BAD-THING' or 'prefer-GOOD-THING' - should this be `no-direct-imports` or `prefer-folder-imports` (or something to that effect)
- How should we enable this rule: Create a new config called "images"? Shoehorn it into polaris config?

### DIFF
--- a/docs/rules/images-no-direct-imports.md
+++ b/docs/rules/images-no-direct-imports.md
@@ -1,0 +1,30 @@
+# Prevent directly importing image files and instead force usage of an index file (image-no-direct-imports)
+
+Image files should live in a directory (e.g. `icons`, `illustrations` or `images`) that must contain a dedicated index file that re-exports all the images in that directory.
+
+Files that consume images must import from that index file instead of importing images directly. This rule enforces that convention by disallowing importing directly from an image in any file except the index file in the same dorectory as the image.
+
+## Rule Details
+
+Example of **incorrect** code for this rule:
+
+
+```js
+// components/Foo/Foo.js
+import icon1 from './icons/icon1.svg';
+```
+
+
+Example of **correct** code for this rule:
+
+```js
+// components/Foo/icons/index.js
+export {default as icon1} from './icon1.svg';
+
+// components/Foo/Foo.js
+import {icon1} from './icons';
+```
+
+## When Not To Use It
+
+If you do not wish to enforce import locations for images, then you can safely disable this rule.
--- a/lib/rules/images-no-direct-imports.js
+++ b/lib/rules/images-no-direct-imports.js
@@ -1,0 +1,56 @@
+const {basename, dirname, extname} = require('path');
+const resolve = require('eslint-module-utils/resolve').default;
+const {docsUrl} = require('../utilities');
+
+function isImageImport(resolvedSource) {
+  return /\.(svg|png|jpg)$/.test(resolvedSource);
+}
+
+function isImportFromCurrentFolderIndex(contextFilename, resolvedSource) {
+  const isIndex =
+    basename(contextFilename, extname(contextFilename)) === 'index';
+
+  return isIndex && dirname(resolvedSource) === dirname(contextFilename);
+}
+
+module.exports = {
+  meta: {
+    docs: {
+      description:
+        'Prefer importing image files from the index file of the directory instead of the direct path to the image file.',
+      category: 'Best Practices',
+      recommended: false,
+      uri: docsUrl('images-no-direct-imports'),
+    },
+    fixable: null,
+  },
+  create(context) {
+    function checkNode(node) {
+      if (!node.source || !node.source.value) {
+        return;
+      }
+
+      const resolvedSource = resolve(node.source.value, context);
+
+      if (
+        resolvedSource &&
+        isImageImport(resolvedSource) &&
+        !isImportFromCurrentFolderIndex(context.getFilename(), resolvedSource)
+      ) {
+        context.report({
+          node,
+          message: `Prefer importing image files from the index file of the directory ("{{folderPath}}") instead of the direct path to the image file ("{{filePath}}").`,
+          data: {
+            folderPath: dirname(node.source.value),
+            filePath: node.source.value,
+          },
+        });
+      }
+    }
+    return {
+      ImportDeclaration: checkNode,
+      ExportNamedDeclaration: checkNode,
+      ExportAllDeclaration: checkNode,
+    };
+  },
+};
--- a/tests/lib/rules/images-no-direct-imports.js
+++ b/tests/lib/rules/images-no-direct-imports.js
@@ -1,0 +1,108 @@
+const {RuleTester} = require('eslint');
+const {fixtureFile} = require('../../utilities');
+const rule = require('../../../lib/rules/images-no-direct-imports');
+
+const ruleTester = new RuleTester({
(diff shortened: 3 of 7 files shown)


docs/rules/images-no-direct-imports.md
```diff
@@ -0,0 +1,30 @@
+# Prevent directly importing image files and instead force usage of an index file
```
cartogram: Add the rule title at the end of this sentence, same as the other docs.

docs/rules/images-no-direct-imports.md
```diff
@@ -0,0 +1,30 @@
...
+
+Image files should live in a directory (e.g. `icons`, `illustrations` or `images`) that must contain a dedicated index file that reexports all the images in that folder.
```
cartogram: `re-exports`

cartogram: How do you feel about this instead for these 2 paragraphs?

>This rule prevents direct imports to image files by enforcing that all images be placed inside a directory (e.g. `icons`, `illustrations` or `images`) that also contains a dedicated index file that re-exports from that folder. Files that consume images must import from that folder instead of importing from the path to the image file directly.

docs/rules/images-no-direct-imports.md
```diff
@@ -0,0 +1,30 @@
...
+// components/Foo/icons/index.js
+export {default as icon1} from './icon1.svg';
+
+// components/Foo/Foo.js
+import {icon1} from './icons';
+```
+
+## When Not To Use It
+
+If you do not wish to enforce imports locations for images, then you can safely disable this rule.
```
cartogram: `imports` => `import`

lib/rules/images-no-direct-imports.js
```diff
@@ -0,0 +1,49 @@
...
+    basename(contextFilename, extname(contextFilename)) === 'index';
+
+  return isIndex && dirname(resolvedSource) === dirname(contextFilename);
+}
+
+module.exports = {
+  meta: {
+    docs: {
+      description:
+        "Prefer that imports of image files go through an index file within the image's directory instead of directly referencing images within a component.",
```
cartogram: A slight re-word maybe? `Prefer importing image files from a within a directory instead of from the path to the image file directly.`

lib/rules/images-no-direct-imports.js
```diff
@@ -0,0 +1,49 @@
...
+
+      if (
+        resolvedSource &&
+        isImageImport(resolvedSource) &&
+        !isImportFromCurrentFolderIndex(context.getFilename(), resolvedSource)
+      ) {
+        context.report({
+          node,
+          message:
+            'Do not import image files directly, instead import from the index of the folder containing the image',
```
cartogram: You are explaining this rule in a few different ways/different wordings. It would be more clear to stick to one.

cartogram (approved):
Code makes sense to me, just made some suggests for the way you describe this rule.

tests/lib/rules/images-no-direct-imports.js
```diff
@@ -0,0 +1,104 @@
...
+    sourceType: 'module',
+  },
+});
+
+function errors(type) {
+  return [
+    {
+      type,
+      message:
+        'Do not import image files directly, instead import from the index of the folder containing the image',
```
cartogram: We don't do it too often in other rules, but you can avoid the repetition here by using a messageId. https://eslint.org/docs/developer-guide/working-with-rules#messageids

cartogram:
> * What's the preferred convention for namespaced rules? A prefix to the filename (e.g. react-type-state) or create a folder (e.g. typescript/prefer-singular-enums)

In this case, I don't think you need a folder prefix. Do we imagine a lot of images related rules? Probably not.

> * Whats the preferred naming between 'no-BAD-THING' or 'prefer-GOOD-THING' - should this be `no-direct-imports` or `prefer-folder-imports` (or something to that effect)

Wish I had a good answer and I imagine we are pretty inconsistent. I think there are official standards out there somewhere but I couldn't find them unfortunately. In the past, I've just chosen the one that sounds right to me 🤷‍♂️ 

In this case I would go with `prefer`, but `images-prefer-no-direct-imports` seems long-winded and I like what you have.

> * How should we enable this rule: Create a new config called "images"? Shoehorn it into polaris config?

I don't think an images config is worthwhile. I think just keep it amongst the other generic ones, this has nothing to do with polaris specifically. If anything, we have a number of file-location-y/import (strict-component-boundaries, no-ancestry-direct-import) rules that would be a more appropriate config-grouping for this.

Also, remember to add this to the Readme, Changelog, etc...

docs/rules/images-no-direct-imports.md
BPScott: I think that implies something slightly different to what is happening. To me that paragraph suggests that is the images that are checked by eslint (e.g. "enforcing all images..."), which is not the case.

The implementation of this rule is slightly roundabout - it don't check anything about the images themselves, it identifies imports to those images and complains if the import is in any file except the index file in the same folder as the image.

BPScott: How about the following:

> Image files should live in a directory (e.g. `icons`, `illustrations` or `images`) that must contain a dedicated index file that re-exports all the images in that folder.
>
> Files that consume images must import from that index file instead of importing images directly. This rule enforces that convention by disallowing importing directly from an image in any file except the index file in the same folder as the image.

BPScott:
>In this case, I don't think you need a folder prefix. Do we imagine a lot of images related rules? Probably not.

There might be one or two more down the line but that's it. No folder it is :)

> In this case I would go with `prefer`, but `images-prefer-no-direct-imports` seems long-winded and I like what you have.

I was thinking that if we do "prefer" then what comes after becomes a positive statement - `images-prefer-folder-import` or `images-prefer-folder-index-import`. but neither of those grab me as being WAY better than `images-no-direct-imports`.

> don't think an images config is worthwhile. I think just keep it amongst the other generic ones, this has nothing to do with polaris specifically

Cool. I'll toss it into the same config as strict-component-boundaries and no-ancestry-direct-import.


Thanks for being a sounding board for my overthinking :)

lib/rules/images-no-direct-imports.js
BPScott: > Prefer importing image files from the index file of the directory instead of the direct path to the image file.

docs/rules/images-no-direct-imports.md
cartogram: That sounds good to me 👍

PR #220, opened 2019-02-12
Simplify a few things that cause weirdness

BPScott:
- all.js is a config that is only used for testing, it has got nothing to do with the rules that get exported, so move it out of lib/config and into the test folder, and update it so its ruleset is built up based on including things from theconfigs we expose rather than importing arbitrary files from lib/config/rules.
- Use eslint-plugin-self in development instead of depending on ourselves, which is weird and has recently caused me some headscratching (and I think is the root cause of some other weirdness that @cartogram has came across in the past)

---

~This is basically good for review, but I won't merge until https://github.com/not-an-aardvark/eslint-plugin-self/pull/2 is merged and released and I've updated this PR to use the version from npm rather than my github fork.~

An updated version of eslint-plugin-self is released and I'm using it now. This is good for review

---

As a follow up to this, now the files in src/config/rules are only mentioned in single files (instead of the config they got used in and all.js) we can look at just putting those objects inline in the files in lib/config to save a level of indirection, but that's some tidying for another day.


cartogram (approved):
This all looks good to me 👍

PR #221, opened 2019-02-13
Adds "every" and "descriptive" to no-vague-titles

cartogram:
When enabling this rule in web it was easy to replace some of the vague words with these. `all` becomes `every`, and `descriptive` is just another filler word. Adding these words to the rule as per @ismail-syed's suggestion.

I also added the `allow` config option to the docs. Closing https://github.com/Shopify/eslint-plugin-shopify/issues/209

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 <!-- ## Unreleased -->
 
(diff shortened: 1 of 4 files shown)


ismail-syed (approved):
Changelog update and this should be 👍

Issue #227, opened 2019-03-09
Rule for no `if` statements in tests


lemonmade:

cartogram:
This should do it: https://astexplorer.net/#/gist/55f61680ccc7e5c4e876d29fe2b3f56f/a55b5d7d16eb2c59974c5dd2e03b5155ad90cc5a, but an alternate approach would be to look for visits on a test statement and then look in the body for `ifStatement`s.

Not sure what would work better TBH.

PR #232, opened 2019-03-22
Adds `jest/no-if`

cartogram:
Closes https://github.com/Shopify/eslint-plugin-shopify/issues/227


docs/rules/jest/no-if.md
```diff
@@ -0,0 +1,34 @@
+# Prevent if statements in tests. (jest/no-if)
+This rule prevents the use of if statements in tests. 
+
+## Rule Details
+
+If statments in tests is usually an indication that a test is attempting to cover too much. Each branch of code executing within an if statement will likely be better served by a test devoted to it.
```
lemonmade: Not only covering too much, but also possibly not actually running the test you think it is

lib/rules/jest/no-if.js
```diff
@@ -0,0 +1,93 @@
+const {docsUrl} = require('../../utilities');
+
+const TEST_FUNCTION_NAMES = [
```
lemonmade: We don't have this list for any other test?

lib/rules/jest/no-if.js
```diff
@@ -0,0 +1,93 @@
...
+  'it',
+  'xit',
+  'fit',
+  'test',
+  'xtest',
+  'describe',
```
lemonmade: I am not sure you need to include `describe` blocks

lib/rules/jest/no-if.js
```diff
@@ -0,0 +1,93 @@
...
+        }
+        inCallExpression = true;
+      },
+      IfStatement(node) {
+        if (!inCallExpression) {
+          return;
+        }
+
+        context.report({
+          messageId: 'noIf',
```
lemonmade, line 43: TIL

lib/rules/jest/no-if.js
```diff
@@ -0,0 +1,93 @@
...
+
+  create(context) {
+    let inCallExpression = false;
+
+    return {
+      CallExpression(node) {
+        if (notTestFunction(node) || hasEmptyBody(node)) {
+          return;
+        }
+        inCallExpression = true;
```
lemonmade: The iffy thing with doing this in scope is that nesting can throw things off. Nesting might not be too much of an issue here but you still might want to go with the more formal mechanism for this (state)

lib/rules/jest/no-if.js
```diff
@@ -0,0 +1,93 @@
...
+        if (!inCallExpression) {
+          return;
+        }
+
+        context.report({
+          messageId: 'noIf',
+          node,
+        });
+      },
+      'CallExpression:exit': () => {
```
lemonmade: I think this probably fails with:

```ts
it('should fail', () => {
  anotherCallExpressionToThrowThingsOff();
  if (ohShoot) {

  }
});
```

lemonmade (changes requested):

yarn.lock
```diff
@@ -3320,10 +3320,10 @@ mem@^1.1.0:
   dependencies:
     mimic-fn "^1.0.0"
 
-merge@1.2.0:
-  version "1.2.0"
-  resolved "https://registry.yarnpkg.com/merge/-/merge-1.2.0.tgz#7531e39d4949c281a66b8c5a6e0265e8b05894da"
-  integrity sha1-dTHjnUlJwoGma4xabgJl6LBYlNo=
+merge@1.2.1:
```
lemonmade, line 3323: Why this change?

lib/rules/jest/no-if.js
cartogram: I figured I'd leave it cause it's not like we want to allow:
```
describe('SomeThing', () => {
  if(!someFlag) {
     it('testing', () => {
     //...
     })
  }
})

```

lib/rules/jest/no-if.js
cartogram: Can you clarify what you mean when you say `more formal mechanism for this (state)`?

Fixing the nesting part should be easy if I just keep track of the depth instead of just a boolean for being in a call expression or not. Maybe that is what you meant by state?

lemonmade: I meant having it on the `context.state` or whatever that field is called so that it is scoped to the part of the tree where you are actually inside the test block (structurally, I don't think it makes much of a difference, but that's what the feature is there for)

lemonmade: Wait actually maybe I am getting mixed up with Babel plugins :P

cartogram: lol you are. I only know because I did the same thing. 😛

lib/rules/jest/no-if.js
cartogram: moved to a require from `utilities/jest`

yarn.lock
cartogram: From here I guess: https://github.com/Shopify/eslint-plugin-shopify/commit/d45d8d07dfabe83e80d48661b235e72c0e123e9e

lemonmade (approved):
👍

PR #233, opened 2019-03-27
Plugin Updates and React Hooks

### LINKED ISSUE #230: Add eslint-plugin-react-hooks

cartogram:
Closes https://github.com/Shopify/eslint-plugin-shopify/issues/230

Updates a bunch of plugins, brings in the `@typescript` packages and adds rules for react-hooks rules.


CHANGELOG.md
```diff
@@ -3,8 +3,53 @@
...
+| `eslint-plugin-prettier` | `3.0.1`  | `4.1.0` |
+
+#### Added Plugins
+
+| Package | Version |
+| ------- | ------- |
+| eslint-plugin-react-hooks | 1.5.0 |
+| @typescript-eslint/eslint-plugin | 1.5.0 |
+| "@typescript-eslint/parser | 1.5.0 |
+| babel-eslint | 10.0.1 |
```
lemonmade, line 65: Personally I am not sure any of these plugins matter, all consumers care about is rules.

CHANGELOG.md
```diff
@@ -3,8 +3,53 @@
...
+
+#### Removed Plugins
+
+| Package |
+| ------- |
+| eslint-plugin-mocha |
+| eslint-plugin-ava |
+| eslint-plugin-flowtype |
+| eslint-plugin-chai-expect |
```
lemonmade, line 74: I think the fact that we have removed all of these is very important to note.

lemonmade: Also need to show the removal of the deprecated configs

lemonmade: And that we made `eslint-comments` part of the default configs

index.js
```diff
@@ -42,24 +42,15 @@ module.exports = {
...
 
     // Augmenting configs - When extending, these go after the core config
-    ava: require('./lib/config/ava'),
-    'eslint-comments': require('./lib/config/eslint-comments'),
-    flow: require('./lib/config/flow'),
     graphql: require('./lib/config/graphql'),
     jest: require('./lib/config/jest'),
     jquery: require('./lib/config/jquery'),
     lodash: require('./lib/config/lodash'),
-    mocha: require('./lib/config/mocha'),
```
lemonmade, line 52: What do you think, should we just go whole-hog and remove more of the ones we basically don't use anymore, like `lodash`/ `jquery`?

lib/config/rules/react-hooks.js
```diff
@@ -0,0 +1,6 @@
+// See https://github.com/facebook/react/tree/master/packages/eslint-plugin-react-hooks
+
+module.exports = {
+  'react-hooks/rules-of-hooks': 'error',
+  'react-hooks/exhaustive-deps': 'error',
```
lemonmade, line 7: No descriptions for either of these?

lemonmade (approved):

index.js
cartogram: I do.

lib/config/react.js
```diff
@@ -11,18 +11,17 @@ module.exports = {
...
   },
 
-  globals: {
-    fetch: true,
-    ReactElement: true,
-    ReactClass: true,
-  },
-
-  plugins: ['react', 'jsx-a11y'],
+  plugins: ['react', 'jsx-a11y', 'react-hooks'],
```
BPScott: We're referencing rules in the shopify plugin below but not enabling it here, should it be present?

lib/config/typescript.js
```diff
@@ -3,11 +3,17 @@ const merge = require('merge');
 module.exports = {
   extends: 'plugin:shopify/esnext',
 
-  plugins: ['typescript'],
+  plugins: [
+    '@typescript-eslint',
+    'eslint-comments',
```
BPScott, line 8: the `typescript` config extends from `esnext`, which in turn extends from `core` which adds the eslint-comments plugin (see https://github.com/Shopify/eslint-plugin-shopify/pull/233/files#diff-03d566093aa3fdf9491a9f677f0fa467R4) so I don't believe you need to specify this again here

BPScott (commented):
Haven't ran the code, but looking at it seems good aside from 2 small oversights. I'm excited to see the addition of react-hooks and simplification by removing rulesets that we no-longer use.

I'll echo Chis's comments about the changelog needing to be improved as it doesn't currently tell people what they need to change to get their project back in working order. The following need to be mentioned:

- The `typescript-prettier` and `typescript-react` rulesets have been removed and what people need to do fix their configs to if they're still using one of those rulesets
- The `eslint-comments` ruleset has been removed and is now enabled by default as part of core - if you're using `es5`, `esnext`, `react` or `typescript` then you can remove the reference to `eslint-comments`.
- The change from `eslint-plugin-typescript` to `@typescript-eslint/eslint-plugin` has resulted in rule names changing and that people will need to update any rules that they define e.g. `typescript/no-var-requires` becomes `@typescript-eslint/no-var-requires`

lib/config/typescript.js
lemonmade: For some reason I did need to enable it again here :(

CHANGELOG.md
```diff
@@ -3,8 +3,53 @@
 <!-- ## Unreleased -->
 ### Added
 
+#### Plugin updates and additions ([#233](https://github.com/Shopify/eslint-plugin-shopify/pull/233))
+
+#### New rules
+
 * `shopify/jest/no-if` ([#232](https://github.com/Shopify/eslint-plugin-shopify/pull/232))
 
+Refer to the [Rules of Hooks documentation](https://reactjs.org/docs/hooks-rules.html) to learn more about the followeing rules.
```
cartogram: `following`

BPScott:
Tangentially related: if we're removing `typescript-react` and `typescript-prettier` then we'll need to update [webgen](https://github.com/Shopify/webgen/blob/master/src/commands/app/files/.eslintrc.js)

README.md
```diff
@@ -62,14 +62,13 @@ If working on an ES5 project, extend the ES5 version of the configuration:
...
-You can also add some "augmenting" configs on top of the "core" config by extending an array of linting configs. For example, this package provides a lodash linting config, which can be added to the ESNext config with the following configuration file:
+You can also add some "augmenting" configs on top of the "core" config by extending an array of linting configs. For example, this package provides a react linting config, which can be added to the ESNext config with the following configuration file:
 
 ```json
 {
   "extends": [
     "plugin:shopify/esnext",
-    "plugin:shopify/lodash",
-    "plugin:shopify/mocha"
+    "plugin:shopify/react",
```
BPScott: `react` extends from `esnext` - so we consider `react` a core config rather than an augmenting one. `jest` might be a better example of an augmenting config.

CHANGELOG.md
```diff
@@ -3,8 +3,81 @@
 <!-- ## Unreleased -->
 ### Added
 
+#### Plugin updates and additions ([#233](https://github.com/Shopify/eslint-plugin-shopify/pull/233))
+
+#### Breaking Changes
+
+* `shopify/jquery-dollar-sign-reference` has been removed.
+
+* The `ava`, `mocha`, `jquery`, `lodash`, `typescript-prettier`, `typescript-react` rulesets have been removed.
```
BPScott: Would it be worth showing what you should use instead of `typescript-prettier` and `typescript-react`?

Something like:

```md
* The `ava`, `mocha`, `jquery` and `lodash` rulesets have been removed as these tools are are not commonly used at Shopify.
* The `typescript-react` and `typescript-prettier` rulesets have been removed. Replace `["plugin:shopify/typescript-react"]` with `["plugin:shopify/typescript", "plugin:shopify/react"]` and replace`["plugin:shopify/typescript-prettier"]`  with `["plugin:shopify/prettier"]`
```

CHANGELOG.md
```diff
@@ -3,8 +3,81 @@
...
+| `eslint-plugin-promise` | `4.0.0` | `4.0.1` |
+| `eslint-plugin-node` | `7.0.1` | `8.0.1` |
+| `eslint-plugin-jsx-a11y` | `6.1.1"`  | `6.2.1` |
+| `eslint-plugin-jest` | `21.22.0`  | `21.23.0` |
+| `eslint-plugin-import` | `2.14.0`  | `2.16.0` |
+| `eslint-plugin-graphql` | `2.1.0-0`  | `3.0.3` |
+| `eslint-plugin-eslint-comments` | `3.0.1`  | `3.1.1` |
+| `eslint-plugin-babel` | `5.1.0`  | `5.3.0` |
+| `eslint-plugin-utils` | `2.1.0`  | `2.3.0` |
+| `eslint-plugin-prettier` | `3.0.1`  | `4.1.0` |
```
BPScott: This didn't happen. The last published version of this is still 3.0.1

BPScott (commented):
A few little wording nitpicks

PR #235, opened 2019-04-09
Remove `describe` from jest/no-if

cartogram:
This PR Removes `describe` from the checked function names of `jest/no-if`.

I tested this rule out in a few projects and if statements in afterEach/ beforeEach blocks and in general within `describe` blocks is a valid thing to do.

### DIFF
--- a/lib/rules/jest/no-if.js
+++ b/lib/rules/jest/no-if.js
@@ -1,8 +1,7 @@
 const {docsUrl} = require('../../utilities');
-const {
-  TEST_FUNCTION_NAMES,
-  getTestMethodName,
-} = require('../../utilities/jest');
+const {getTestMethodName} = require('../../utilities/jest');
+
+const TEST_FUNCTION_NAMES = ['it', 'xit', 'fit', 'test', 'xtest'];
 
 module.exports = {
   meta: {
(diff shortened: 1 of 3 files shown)


lib/rules/jest/no-if.js
```diff
@@ -1,8 +1,7 @@
 const {docsUrl} = require('../../utilities');
-const {
-  TEST_FUNCTION_NAMES,
-  getTestMethodName,
-} = require('../../utilities/jest');
+const {getTestMethodName} = require('../../utilities/jest');
+
+const TEST_FUNCTION_NAMES = ['it', 'xit', 'fit', 'test', 'xtest'];
```
cartogram, line 4: I think the clearest thing is to just redefine these locally for this test.

yarn.lock
```diff
@@ -1761,13 +1761,6 @@ eslint-plugin-jsx-a11y@6.2.1:
     has "^1.0.3"
     jsx-ast-utils "^2.0.1"
 
-eslint-plugin-lodash@5.1.0:
```
cartogram, line 1764: This is a mistake leftover from a previous pr.

tests/lib/rules/jest/no-if.js
```diff
@@ -122,9 +123,9 @@ ruleTester.run('no-if', rule, {
       ],
     },
     {
-      code: `fdescribe.only('foo', () => {
-        if('bar') {}
-      })`,
+      code: `fit('foo', () => {
+          if('bar') {}
+        })`,
```
patsissons: extra space?

patsissons (approved):
LGTM

PR #237, opened 2019-04-09
Adds `hooks-strict-return`

### LINKED ISSUE #234: Add lint rule for "what do I return from a hook?"
Or, just a best practice guide, if we can't do a linting rule.

cartogram:
closes https://github.com/Shopify/eslint-plugin-shopify/issues/234

### DIFF
--- a/lib/rules/react-hooks-strict-return.js
+++ b/lib/rules/react-hooks-strict-return.js
@@ -17,7 +17,7 @@ module.exports = {
   },
 
   create(context) {
-    let inHook = false;
+    let inHook = 0;
 
     return {
       VariableDeclarator(node) {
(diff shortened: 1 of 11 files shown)


cartogram:
@lemonmade @GoodForOneFare @michelleyschen @TheMallen Can I get some initial feedback on this rule? Is this what we want to enforce? Anything I am missing?

Worth noting that I am not looking to get this to the point where it will know everything about what you are returning. For example, if the variables you are returning are the result of a function call or from another file.

lemonmade:
I am board with this 👍

lib/rules/react-hooks-strict-return.js
```diff
@@ -0,0 +1,154 @@
...
+        }
+
+        inHook = true;
+      },
+      'FunctionDeclaration:exit': function(node) {
+        if (!isHook(node)) {
+          return;
+        }
+
+        inHook = false;
```
lemonmade: I suspect something like this will cause the rule to not complain, even though it should:

```tsx
function useFoo() {
  useEffect(() => {});

  return [1, 2, 3, 4];
}
```

lemonmade (approved):
Code LGTM but I think the flow control bits are probably going to prevent some violations from being caught

lib/rules/react-hooks-strict-return.js
cartogram: Added some more tests to cover this https://github.com/Shopify/eslint-plugin-shopify/pull/237/files#diff-cde867597099df6969da77e4a149d359R206

PR #238, opened 2019-04-09
Fix duplicate `camelcase` rule

cartogram:
Currently, we are getting duplicate reports for camelcase variables.

To solve this in this PR I am keeping the `babel/camelcase` rule on unless they are using the typescript config, at which point we can use the `@typescript-eslint/camelcase` rule and turn off the babel rule.

Screen Shot 2019-04-09 at 10 08 30 AM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 <!-- ## Unreleased -->
 
+### Removed
+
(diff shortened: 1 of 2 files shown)


GoodForOneFare (approved):
👍 Makes sense.  It'd be nice if this was covered by `tests/lib/config/typescript.js`, but setting up a test is probably more effort than it's worth.

CHANGELOG and 🚢 👍

Issue #231, opened 2019-03-21
Make sure React linting rules are appropriate for hooks


lemonmade:
In particular, we must allow arrow functions in JSX attributes, since methods will not be an option, and you need to close over variables fairly often.

lemonmade:
@cartogram this one down now?

PR #239, opened 2019-04-09
Enable jsx-no-bind

cartogram:
This PR changes the config for `react/jsx-no-bind` to allow functions as part of https://github.com/Shopify/eslint-plugin-shopify/issues/231.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 <!-- ## Unreleased -->
 
+### Changed
+
+* `react/jsx-no-bind` now allows function and arrow functions. ([#239](https://github.com/Shopify/eslint-plugin-shopify/pull/239))
+
 ### Removed
 
 * turned off `babel/camelcase` in typescript config because it overlaps with `@typescript-eslint/camelcase`. ([#238](https://github.com/Shopify/eslint-plugin-shopify/pull/238))
--- a/lib/config/rules/react.js
+++ b/lib/config/rules/react.js
@@ -128,8 +128,9 @@ module.exports = {
   'react/jsx-max-depth': 'off',
   // Limit maximum of props on a single line in JSX
   'react/jsx-max-props-per-line': 'off',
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 <!-- ## Unreleased -->
 
+### Changed
+
+* `react/jsx-no-bind` now allows function and arrow functions. ([#239](https://github.com/Shopify/eslint-plugin-shopify/pull/239))
 ### Removed
```
lemonmade, line 9: Newline above

lib/config/rules/react.js
```diff
@@ -128,8 +128,14 @@ module.exports = {
...
   // Limit maximum of props on a single line in JSX
   'react/jsx-max-props-per-line': 'off',
-  // Prevent usage of .bind() and arrow functions in JSX props
-  'react/jsx-no-bind': 'error',
+  // Prevent usage of .bind() in JSX props
+  'react/jsx-no-bind': [
+    'error',
+    {
+      allowArrowFunctions: true,
+      allowFunctions: true,
```
lemonmade: What does this mean is not allowed?

lemonmade (approved):

lib/config/rules/react.js
cartogram: only `bind` calls are not allowed. The other config options have to do with ignoring refs/dom elements.

cartogram: And they default to false:

```
  "ignoreDOMComponents": <boolean> || false,
  "ignoreRefs": <boolean> || false,
  "allowArrowFunctions": <boolean> || false,
  "allowFunctions": <boolean> || false,
  "allowBind": <boolean> || false
```

lemonmade: I think we may as well allow bind calls then, there's no real difference of `this.handleFoo.bind(this)` and `() => this.handleFoo()`

PR #243, opened 2019-04-18
Fix restrict-full-import

### LINKED ISSUE #222: restrict-full-import fails on "empty" array pattern
Failing code:

```js
const [, foo] = bar;
```

we try to access `type`, but the empty slot is actually just `null`:

screen shot 2019-02-15 at 11 53 56 am

cartogram:
Closes https://github.com/Shopify/eslint-plugin-shopify/issues/222

### DIFF
--- a/lib/rules/restrict-full-import.js
+++ b/lib/rules/restrict-full-import.js
@@ -41,7 +41,7 @@ module.exports = {
           })) ||
         (left.type === 'ArrayPattern' &&
           left.elements.some((element) => {
-            return element.type === 'RestElement';
+            return element != null && element.type === 'RestElement';
           }))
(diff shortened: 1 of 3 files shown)


lib/rules/restrict-full-import.js
```diff
@@ -41,6 +41,9 @@ module.exports = {
           })) ||
         (left.type === 'ArrayPattern' &&
           left.elements.some((element) => {
+            if (element === null) {
+              return false;
+            }
             return element.type === 'RestElement';
```
lemonmade: I’d probably just do element != null && here

lemonmade (approved):

PR #247, opened 2019-04-24
Optimize images-no-direct-imports rule

BPScott:
Check the extension of the import before trying to resolve it. This
allows us to escape early and not trigger trying to resolve the path
(which is expensive compared to string comparisons) if the import isn't
for an image.


cartogram (approved):
💯

PR #249, opened 2019-04-26
Revert Typescript update to fix prettier

cartogram:
This PR reverts the typescript updates from https://github.com/Shopify/eslint-plugin-shopify/pull/233 because they break the vscode prettier plugin. Without these changes, vscode will not format code using prettier-eslint. The following error will display in the output:

```
Error while loading rule '@typescript-eslint/no-unnecessary-qualifier'/Users/mattseccafien/src/github.com/Shopify/graphql-tools-web/src/utilities.ts:: You have used a rule which requires parserServices to be generated. You must therefore provide a value for the "parserOptions.project" property for @typescript-eslint/parser.
Occurred while linting /Users/mattseccafien/src/github.com/Shopify/graphql-tools-web/src/utilities.ts
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,15 +65,13 @@
     "prettier": "^1.17.0",
     "react": "^16.8.6",
     "react-dom": "^16.8.6",
-    "typescript": "^3.4.3"
+    "typescript": "~3.1.1"
   },
   "peerDependencies": {
     "eslint": "<6 >5.0.0",
     "prettier": "^1.14.0"
   },
   "dependencies": {
-    "eslint-plugin-typescript": "0.12.0",
-    "typescript-eslint-parser": "20.0.0",
     "babel-eslint": "10.0.1",
     "common-tags": "^1.8.0",
     "eslint-config-prettier": "4.1.0",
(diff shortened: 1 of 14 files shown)


package.json
```diff
@@ -91,9 +89,11 @@
     "eslint-plugin-react": "7.12.4",
     "eslint-plugin-react-hooks": "^1.6.0",
     "eslint-plugin-sort-class-members": "1.4.0",
+    "eslint-plugin-typescript": "0.12.0",
     "merge": "1.2.1",
     "pascal-case": "^2.0.1",
     "pkg-dir": "4.1.0",
-    "pluralize": "^7.0.0"
+    "pluralize": "^7.0.0",
+    "typescript-eslint-parser": "20.0.0"
```
lemonmade, line 97: Do you know what versions of TypeScript this supports?

tests/fixtures/typescript-imports/.eslintrc.js
```diff
@@ -1,8 +1,8 @@
...
   plugins: ['self'],
   extends: ['plugin:self/typescript', 'plugin:self/node'],
-  parserOptions: {
-    project: 'tsconfig.json',
-    tsconfigRootDir: __dirname,
-  },
+  // parserOptions: {
+  //   project: 'tsconfig.json',
+  //   tsconfigRootDir: __dirname,
+  // },
```
lemonmade: 🔥

lemonmade (approved):

package.json
cartogram: Yeah it should be on `~3.2.1`, thanks for catching that.

cartogram: Actually, what I had was right. `typescript-estree` doesn't support past `~3.1.1`.

PR #258, opened 2019-05-06
Fix `react-prefer-private-members` nested classes

### LINKED ISSUE #255: `shopify/react-prefer-private-members` warns for members in wrong class when the react class is nested inside another one
I've updated to the latest config (v28) and the rule `shopify/react-prefer-private-members` warn for a member in the wrong class. This seems to happen when a React class is defined inside another class.

For shopifolks, [more context is available here](https://shopify.slack.com/archives/CCNRSKLTH/p1556792873155900).

This can be reproduced by defining a React class in the constructor of another class (not especially react class). All members from the top class will warns that the member is not private in the nested class.

cartogram:
closes https://github.com/Shopify/eslint-plugin-shopify/issues/255

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 - New Rule: `jest/no-mocks-import`. This rule disallows manually importing from `__mocks__`.
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


tleunen (approved):
Thanks for the quick fix! :)

PR #262, opened 2019-05-12
Adds `no-namespaced-import` rule

cartogram:
This PR adds a new rule to prevent namespace imports.

#### ToDo:

- [x] Add Fixer
- [x] Add options for additional modules and ignored/whitelisted modules
- [x] Add documentation
- [x] Add configurations
- [x] More tests probably
- [x] Add to Changelog

#### Reviewers:

I'd like reviewers to focus on the following:
1) The description of the rule. Does it communicate enough about why we don't want namespace imports? Should we include any information about typescript specifically (--esModuleInterop, etc...).
1) The fixer for the rule. Converts namespace to default imports, which is not always the solution, for example `import * as foo from 'Foo'` might actually want to be `import {foo, bar, baz} from 'Foo'`. The fixer could also lead to breaking code.
1) Is the `allow` option. Is this enough configuration or should we an an array option to match modules to report (the opposite of allow).


GoodForOneFare:
Approach LGTM.

docs/rules/no-namespace-imports.md
```diff
@@ -0,0 +1,46 @@
+# Prevent namespace import declarations. (no-namespace-imports)
+
+Namespace imports grab all exported variables from a module as nested properties under a single variable name. This approach of importing everything, rather than only what is required from the module, leads to a polluted local scope from the unused components that were brought in under the module namespace.
```
lemonmade: It does not polite the local scope, that's half the benefit of namespace imports. If we're going to sell this on anything, it should be that this style of import breaks tree shaking.

docs/rules/no-namespace-imports.md
```diff
@@ -0,0 +1,46 @@
...
+
+import * as React from 'react';
+
+```
+
+Examples of **correct** code for this rule:
+
+```ts
+
+import React from 'react';
```
lemonmade: I'd rather show an example showing named imports > namespace import. Yes, this rule does help us enforce the right style for React, but that's something that could conceivably be addressed in types instead.

docs/rules/no-namespace-imports.md
```diff
@@ -0,0 +1,46 @@
...
+```
+
+### `allow`
+
+```json
+{
+  "shopify/no-namespace-imports": [
+    "error",
+    {
+      "allow": ["react", "some-custom-module"]
```
lemonmade: Probably best not to show it for React, we'd never want them to do that, would we?

lib/rules/no-namespace-imports.js
```diff
@@ -0,0 +1,83 @@
...
+    const allowed = options.allow || [];
+
+    function report(node) {
+      const name = node.source.value;
+
+      context.report({
+        node,
+        messageId: 'namespaceImport',
+        data: {name},
+        fix: (fixer) => {
```
lemonmade: I don't think it's safe to provide a fixer, we have no idea what members they reference, and turning it into a default import is not going to be correct in many cases.

lib/rules/no-namespace-imports.js
```diff
@@ -0,0 +1,83 @@
...
+
+function getNamespaceSpecifiers(node) {
+  return node.specifiers.filter((specifier) => {
+    return specifier.type === 'ImportNamespaceSpecifier';
+  });
+}
+
+function ignoreModule(node, allowed) {
+  return allowed.some((stringOrRegex) => {
+    const regex = new RegExp(stringOrRegex);
```
lemonmade: Better to construct this regex array upfront.

lemonmade (changes requested):

cartogram:
superseded by https://github.com/Shopify/eslint-plugin-shopify/pull/305

cartogram:
After trying the import/no-namespace rule, I do think we want a custom rule after all. I'll continue with this work and get it ready for another review.

PR #265, opened 2019-05-15
adds necessary to `shopify/jest/no-vague-terms`

cartogram:
🍑

### DIFF
--- a/tests/lib/rules/jest/no-vague-titles.js
+++ b/tests/lib/rules/jest/no-vague-titles.js
@@ -142,7 +142,53 @@ ruleTester.run('no-vague-titles', rule, {
   ],
   invalid: [
     {
-      code: "it('properly should correcly appropriate all')",
+      code:
+        "it('properly should correcly appropriate all descriptive necessary')",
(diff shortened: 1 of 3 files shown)


tests/lib/rules/jest/no-vague-titles.js
```diff
@@ -142,7 +142,53 @@ ruleTester.run('no-vague-titles', rule, {
   ],
   invalid: [
     {
-      code: "it('properly should correcly appropriate all')",
+      code:
+        "it('properly should correcly appropriate all descriptive necessary')",
```
GoodForOneFare, line 146: Why the word salad? 🤔

GoodForOneFare (approved):
Enforcing my nitpicks org wide!  BWA-HA-HA

tests/lib/rules/jest/no-vague-titles.js
cartogram: Not sure what my original thinking was... probably that they all error together are listed properly in the error message. Going to leave it for now.

PR #298, opened 2019-06-08
Adds conditional statements to `shopify/jest/no-if`

### LINKED ISSUE #296: [jest/no-if] You can bypass this with ternary statements
It'd be nice to add detection for ternary statements as well.

cartogram:
Closes https://github.com/Shopify/eslint-plugin-shopify/issues/296

### DIFF
--- a/lib/rules/jest/no-if.js
+++ b/lib/rules/jest/no-if.js
@@ -20,6 +20,14 @@ module.exports = {
         'Consider breaking the if statement out',
         'into a separate test to resolve this error.',
       ].join(' '),
+      noConditional: [
+        'Tests should not contain conditional statements.',
+        'This is usually an indication that you',
+        'are attempting to test too much at once',
+        'or not testing what you intend to.',
+        'Consider breaking the conditional statement out',
+        'into a separate test to resolve this error.',
+      ].join(' '),
     },
   },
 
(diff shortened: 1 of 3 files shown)


lib/rules/jest/no-if.js
```diff
@@ -20,6 +20,14 @@ module.exports = {
         'Consider breaking the if statement out',
         'into a separate test to resolve this error.',
       ].join(' '),
+      noConditional: [
+        'Tests should not contain conditional statements.',
+        'This is usually an indication that you',
+        'are attempting to test too much at once',
+        'or not testing what you intend to.',
```
lemonmade, line 27: This part of the sentence doesn't read quite right

lib/rules/jest/no-if.js
```diff
@@ -43,6 +51,16 @@ module.exports = {
           node,
         });
       },
+      ConditionalExpression(node) {
```
lemonmade, line 54: One other change I'd really like to see (I think I opened a separate issue for it) is not to complain about this in nested functions. IMO, this is totally fine:

```tsx
it('works', () => {
  const graphQL = createGraphQL({variables} => variables.first ? listOf(variables.first) : listOf(50));
});
```

lemonmade (approved):

PR #299, opened 2019-06-08
Fixes `shopify/jest/no-if` to now ignore if statements inside block statements 

### LINKED ISSUE #263: jest/no-if should not complain when not directly nested in `it`
The following code currently warns, but shouldn't:

```ts
it('works', () => {
  const values = something.map((thing) => {
    if (thing.isFoo) {
      return thing.foo
    } else {
      return thing.bar;
    }
  });

  expect(values).toStrictEqual(['foo']);
});
```

cartogram:
Closes https://github.com/Shopify/eslint-plugin-shopify/issues/263

### DIFF
--- a/lib/rules/jest/no-if.js
+++ b/lib/rules/jest/no-if.js
@@ -30,62 +30,35 @@ module.exports = {
   },
 
   create(context) {
-    const state = {
-      testDepth: 0,
-      nestedDepth: 0,
-    };
+    const stack = [];
 
-    const changeNestedDepth = (increase) => () =>
-      inTest(state) && (increase ? state.nestedDepth++ : state.nestedDepth--);
+    function validate(node) {
+      const lastCallExpression = stack[stack.length - 1];
 
-    const changeDepth = (increase) => (node) =>
-      !testCallExpression(node) &&
-      (increase ? state.testDepth++ : state.testDepth--);
+      if (!lastCallExpression) {
+        return;
+      }
 
-    const ifStatement = (messageId) => (node) =>
-      !ignoreStatement(state) &&
-      context.report({
-        messageId,
-        node,
-      });
+      if (isTestDefinition(lastCallExpression)) {
+        const messageId =
+          node.type === 'ConditionalExpression' ? 'noConditional' : 'noIf';
+
+        context.report({
+          messageId,
+          node,
+        });
+      }
+    }
 
     return {
-      CallExpression: changeDepth(true),
-      FunctionExpression: changeNestedDepth(true),
-      ArrowFunctionExpression: changeNestedDepth(true),
-      IfStatement: ifStatement('noIf'),
-      ConditionalExpression: ifStatement('noConditional'),
-      'FunctionExpression:exit': changeNestedDepth(),
-      'ArrowFunctionExpression:exit': changeNestedDepth(),
-      'CallExpression:exit': changeDepth(),
+      CallExpression(node) {
+        stack.push(node);
+      },
+      IfStatement: validate,
+      ConditionalExpression: validate,
+      'CallExpression:exit': function() {
+        stack.pop();
+      },
     };
   },
 };
-
-function inTest(state) {
-  return state.testDepth > 0;
-}
-
-function ignoreStatement(state) {
-  return state.testDepth === 0 || state.nestedDepth > state.testDepth;
-}
-
-function testCallExpression(node) {
-  return notTestFunction(node) || hasEmptyBody(node);
-}
-
-function notTestFunction(node) {
-  const method = getTestMethodName(node);
-  return !TEST_FUNCTION_NAMES.includes(method);
-}
-
-function hasEmptyBody(node) {
-  return (
-    node.arguments &&
-    node.arguments.length === 2 &&
-    node.arguments[1].type === 'ArrowFunctionExpression' &&
-    node.arguments[1].body &&
-    node.arguments[1].body.body &&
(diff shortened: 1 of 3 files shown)


lib/rules/jest/no-if.js
```diff
@@ -94,3 +106,10 @@ function hasEmptyBody(node) {
     node.arguments[1].body.body.length === 0
   );
 }
+
+function skip(state) {
+  return (
+    state.callExpressionDepth === 0 ||
+    state.blockStatementDepth > state.callExpressionDepth
```
lemonmade: I have a little trouble understanding this condition, can you explain it?

lemonmade (commented):
lol my other review looks really dumb now

lib/rules/jest/no-if.js
cartogram: So the first part is the same as what was there previously if we are not inside of a testing-CallExpression (`it()`, `test()`, etc...), then we don't care about if statements. 

The part after the `||` is new and is trying to determine if we are inside of a BlockStatement that is inside of (but not attached to) a testing-CallExpesssion. Because the opening `{...}` to the CallExpesssion, for example in `it('works'){ //here... }`, will increase both our blockStatementDepth and callExpessionDepth by one, the greater-than check will tell us if we have entered a BlockStatement that is NOT attached to a CallExpression. In other words, if the blockStatementDepth is greater than the callStatementDepth, than we have entered at least one more BlockStatment, while also inside of a testing-CallExpession, which is valid and we want to `skip` (not report) the current ifStatement.

Open to other ways to handle this as well!

lemonmade: Does it catch things like defining an in-line function in an it block (const foo = () => bar ? baz : qux)?

cartogram: I think that should be covered already, will add a test and confirm 👍

cartogram: @lemonmade added passing tests for the arrow functions as well.

lib/rules/jest/no-if.js
```diff
@@ -25,63 +25,60 @@ module.exports = {
         'This is usually an indication that you',
         'are attempting to test too much at once',
         'or not testing what you intend to.',
-        'Consider breaking the conditional statement out',
-        'into a separate test to resolve this error.',
+        'Consider writing a separate test for',
+        'each fork in the conditional statement.',
```
lemonmade, line 30: The other case I often see this: people have a possibly null value and bail out to not make TypeScript mad. We should call that one out, and push them to use `!` instead

lib/rules/jest/no-if.js
```diff
@@ -25,63 +25,60 @@ module.exports = {
...
-          return;
-        }
-        callExpressionDepth++;
-      },
-      IfStatement(node) {
-        if (callExpressionDepth === 0) {
-          return;
-        }
+    const changeNestedDepth = (increase) => () =>
+      inTest(state) && (increase ? state.nestedDepth++ : state.nestedDepth--);
```
lemonmade: This is too clever by a lot IMO, add some blocks and conditions instead

lib/rules/jest/no-if.js
```diff
@@ -25,63 +25,60 @@ module.exports = {
...
+
+    return {
+      CallExpression: changeDepth(true),
+      FunctionExpression: changeNestedDepth(true),
+      ArrowFunctionExpression: changeNestedDepth(true),
+      IfStatement: ifStatement('noIf'),
+      ConditionalExpression: ifStatement('noConditional'),
+      'FunctionExpression:exit': changeNestedDepth(),
+      'ArrowFunctionExpression:exit': changeNestedDepth(),
+      'CallExpression:exit': changeDepth(),
```
lemonmade: Not sure if me or the pattern, but I had to read through this a lot to understand what the `changeNestedDepth`/ `changeDepth`/ with/ without argument actually meant.

lemonmade: Personally, I'd probably just keep a stack of call expressions (as an array), every time you enter push `true` or `false` for if it's a test or not, every time you exit pop. Then, checking for an invalid conditional is just checking if the current stack entry is `true` or `false`.

lemonmade (changes requested):

cartogram:
@lemonmade ready for another look

lib/rules/jest/no-if.js
```diff
@@ -17,69 +17,54 @@ module.exports = {
...
         'This is usually an indication that you',
         'are attempting to test too much at once',
         'or not testing what you intend to.',
-        'Consider breaking the conditional statement out',
-        'into a separate test to resolve this error.',
+        'Consider writing a separate test for',
+        'each fork in the conditional statement.',
+        'If your conditionals are required to',
+        'satisfy the typescript type checker, consider',
+        'using a non-null assertion operator (!) instead.',
```
lemonmade, line 33: ❤️

lib/rules/jest/no-if.js
```diff
@@ -17,69 +17,54 @@ module.exports = {
...
-      },
-      ConditionalExpression(node) {
-        if (callExpressionDepth === 0) {
-          return;
-        }
+      if (!lastCallExpression) {
+        return;
+      }
+
+      if (isTestDefinition(lastCallExpression)) {
```
lemonmade: Is there a reason to compute this manually every time? Can the stack just have the results of calling this?

lemonmade (approved):

PR #300, opened 2019-06-11
Prefer `toThrow` over try/catch/expect via `jest/no-try-expect`

GoodForOneFare:
### What?
Suggest that developers use `toThrow` in tests (see my PR in web for real world examples of why this is helpful 😄).

I really wanted this to be an all-singing, all dancing rule.  In an ideal world, it would give specific feedback for sync vs async functions.  And autofix simple cases.  Ohwell, this is a start 😆

### DIFF
--- a/docs/rules/jest/no-try-expect.md
+++ b/docs/rules/jest/no-try-expect.md
@@ -1,0 +1,42 @@
+# Prevent catch assertions in tests. (jest/no-if)
+This rule prevents the use of `expect` inside `catch` blocks.
+
+## Rule Details
+
+Expectations inside a `catch` block can be silently skipped. While Jest provides an `expect.assertions(number)` helper, Shopify rarely uses this feature. Using `toThrow` concisely guarantees that an exception was thrown, and that its contents match expectations.
+
+Examples of **incorrect** code for this rule:
+
+```js
+it('foo', () => {
+  try {
+    foo(); // `foo` may be refactored to not throw exceptions, yet still appears to be tested here.
+  } catch (err) {
+    expect(err).toMatch(/foo error/);
+  }
+})
+
+it('bar', async () => {
+  try {
+    await foo();
+  } catch (err) {
+    expect(err).toMatch(/foo error/);
+  }
+})
+```
+
+Examples of **correct** code for this rule:
+
+```js
+it('foo', () => {
+  expect(() => foo()).toThrow(/foo error/);
+})
+
+it('bar', async () => {
+  await expect(fooPromise).rejects.toThrow(/foo error/);
+})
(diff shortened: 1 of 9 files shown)


docs/rules/jest/no-try-expect.md
```diff
@@ -0,0 +1,42 @@
+# Prevent catch assertions in tests. (jest/no-if)
+This rule prevents the use of `expect` inside `catch` blocks.
+
+## Rule Details
+
+Expectations inside a `catch` block can be silently skipped.  While Jest provides an `expect.assertions(number)` helper, Shopify rarely uses this feature.  Using `toThrow` concisely guarantees that an exception was thrown, and that its contents match expectations.
```
cartogram: One space is plenty 😛 

```suggestion
Expectations inside a `catch` block can be silently skipped. While Jest provides an `expect.assertions(number)` helper, Shopify rarely uses this feature. Using `toThrow` concisely guarantees that an exception was thrown, and that its contents match expectations.
```

docs/rules/jest/no-try-expect.md
```diff
@@ -0,0 +1,42 @@
...
+```js
+it('foo', () => {
+  try {
+    foo(); // `foo` may be refactored to not throw exceptions, yet still appears to be tested here.
+  } catch (err) {
+    expect(err).toMatch(/foo error/);
+  }
+})
+
+it('bar', () => {
```
cartogram:
```suggestion
it('bar', async () => {
```

lib/rules/jest/no-try-expect.js
```diff
@@ -0,0 +1,61 @@
...
+      'CallExpression:exit': (node) => {
+        if (isTestDefinition(node)) {
+          isTest = false;
+        }
+      },
+    };
+  },
+};
+
+function isTestDefinition(node) {
```
cartogram: What do you think about moving this to `utilities/jest` and importing it from there? That feels nicer than exporting/importing the constant `TEST_FUNCTION_NAMES`. 

We can reuse this utility in `jest/no-if` (can be something done in a later PR).

cartogram (approved):

cartogram:
LGTM, just a few minor comments. You can always add a fixer later on 🙂

lib/rules/jest/no-try-expect.js
GoodForOneFare: Good idea 👍 Done!

GoodForOneFare:
Added boilerplatey config/rules setup + CHANGELOG.  Tested in web, and I'm seeing the lone instance of this rule flagged 🎉

PR #305, opened 2019-06-17
Enable import/no-namespace

cartogram:
After going through the process of writing a [first-party, opinionated version of this rule](https://github.com/Shopify/eslint-plugin-shopify/pull/305), I have settled upon just using the third-party rule from `eslint-plugin-import` instead.


GoodForOneFare (approved):
👍 Off the top of my head, a couple of libs in web _needed_ namespaced imports, but that's what `eslint-disable` is for.

PR #316, opened 2019-06-20
Forbid `I` prefixes in interface names

GoodForOneFare:
### Why tho?
This [TypeScript handbook](https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#names) says not to (it also says that it's only a guideline for contribution to TypeScript itself so...) [this StackOverflow answer](https://stackoverflow.com/questions/31876947/confused-about-the-interface-and-class-coding-guidelines-for-typescript/41967120#41967120) is pretty convincing.

Testing this in an internal project shows a few hits, but it's a low amount of effort to manually fix them:

web (zsh) 2019-06-20 16-49-22

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # Changelog
 
-<!-- ## Unreleased -->
+## Unreleased
+
+### Changed
+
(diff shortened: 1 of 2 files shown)


cartogram (approved):
Too easy

sumo

PR #319, opened 2019-06-24
[WIP] Adds `react-import-first` rule

cartogram:
Part of https://github.com/Shopify/web-foundation/issues/109

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+### New Rules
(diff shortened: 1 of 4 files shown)


BPScott:
Is there some way we can teach the import/order rule a way to order explictly named imports instead of just contextual groups? Otherwise we might end up with these rules conflicting if we have a builtin function and React imported in the same file.

https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md

cartogram:
@BPScott you are super right, I don't think this is the way to go here.

PR #330, opened 2019-07-02
Support eslint v6

cartogram:
Brings us to eslint@6.x. Not having this update is causing some dependabot PRs to fail.

Update tests 'parser' requiring an absolute path rather than a package name. Details found in the [migration guide](https://eslint.org/docs/6.0.0/user-guide/migrating-to-6.0.0#rule-tester-parser). 

>In ESLint v6.0.0, RuleTester disallows parser property with a package name. To address: If you use parser property with package names in test cases, update it with require.resolve() function to resolve the package name to the absolute path to the package.

Without these changes to the tests, these tests expectedly failed due to the parser option not being an absolute path; with the change to the tests, they all pass.

### DIFF
--- a/lib/config/rules/possible-errors.js
+++ b/lib/config/rules/possible-errors.js
@@ -6,26 +6,15 @@ module.exports = {
   // Enforce return statements in getters
   'getter-return': ['error', {allowImplicit: true}],
   // disallow using an async function as a Promise executor
-  'no-async-promise-executor': 'off',
+  'no-async-promise-executor': 'error',
   // Disallow await inside of loops
   'no-await-in-loop': 'off',
   // Disallow comparing against -0
   'no-compare-neg-zero': 'error',
-  // Disallow or enforce trailing commas
-  'comma-dangle': [
-    'error',
-    {
-      arrays: 'always-multiline',
-      objects: 'always-multiline',
-      imports: 'always-multiline',
-      exports: 'always-multiline',
-      functions: 'always-multiline',
-    },
-  ],
   // Disallow assignment in conditional expressions
   'no-cond-assign': 'error',
   // Disallow use of console
-  'no-console': 'error',
+  'no-console': 'off',
   // Disallow use of constant expressions in conditions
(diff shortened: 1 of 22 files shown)


tests/lib/rules/webpack/no-unnamed-dynamic-imports.js
```diff
@@ -1,91 +1,84 @@
...
-      return System.import<{default: any}>(
-        /* webpackChunkName: "bar" */ './bar',
-      ).then(bar => bar);
-    }
-    `,
-    parser: 'typescript-eslint-parser',
-  },
-];
-
-const invalidExamples = [
```
cartogram, line 63: only changed for consistency with the rest of this repo: we do not ever store the examples in a variable before passing them into the ruleTester.

GoodForOneFare (approved):
👍 Adding this to CHANGELOG would be great.  

Scanning through the migration guide, maybe we should enable `no-async-promise-executor`, as it's now recommended?  That doesn't have to happen here, though.

lib/config/rules/possible-errors.js
```diff
@@ -6,26 +6,15 @@ module.exports = {
...
-      imports: 'always-multiline',
-      exports: 'always-multiline',
-      functions: 'always-multiline',
-    },
-  ],
   // Disallow assignment in conditional expressions
   'no-cond-assign': 'error',
   // Disallow use of console
-  'no-console': 'error',
+  'no-console': 'off',
```
GoodForOneFare, line 17: I have a, "some men just want to watch the world burn" fascination for seeing how this plays out... _but_, if we're following the spirit of the migration guide:
- This should still be enabled by default
- It should be disabled for [the node config](https://github.com/Shopify/eslint-plugin-shopify/blob/master/lib/config/node.js)

cartogram: 😬 https://github.com/Shopify/eslint-plugin-shopify/pull/332

PR #331, opened 2019-07-03
Don't report nested arrow functions in `jest/no-if`

### LINKED ISSUE #311: False positive in shopify/jest/no-if
One more edge case relating to #263.

It warns when the if is within a function, such as a react component you're defining within a test:

```
it('has a child with contentContext', () => {
    const Child: React.SFC = (_props, context) => {
      return context.withinContentContainer ? <div /> : null;
    };
}
```

As discovered in https://github.com/Shopify/polaris-react/pull/1713

cartogram:
Fixes https://github.com/Shopify/eslint-plugin-shopify/issues/311

This PR changes jest/no-if so that it reports on if statements that are nested in arrow functions (all of which would be nested in a test CallExpression). This fix was done in 2 steps:
1) Push ArrowFunctions on to our in-memory stack
2) Store those items as a boolean, true if it is part of the test structure (the second argument to it/test/etc...)

The rest of the logic remained the same. We check to see if the last item pushed was true/false, and report if it is.

In other words, "if we are an if statement and we are just inside an arrow function and that arrow function that we are just inside is part of an it/test/etc... function, then we report it".

I implemented this by adding a new jest utility for checking if an ArrowFunction is part of an it/test/etc... function.

cc @BPScott

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 - Enabled `typescript/interface-name-prefix` to prevent `I` prefixes in TypeScript interface names
 
 ### Fixed
-- [Patch] Fix `jest/no-if` from falsely reporting if statements inside of arrow functions ([331](https://github.com/Shopify/eslint-plugin-shopify/pull/331))
+- [Patch] Fix `jest/no-if` from falsely reporting if statements inside of functions ([331](https://github.com/Shopify/eslint-plugin-shopify/pull/331))
 
 ## [29.0.2] - 2019-06-18
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,9 +3,11 @@
 ## Unreleased
 
 ### Changed
-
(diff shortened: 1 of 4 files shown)


BPScott:
Can we identify regular functions in addition to arrow function?

The following should all be valid:

```js
{
      code: `it('foo', () => {
        const foo = function(bar) {
          return foo ? bar : null;
        };
      });`,
    },
    {
      code: `it('foo', () => {
        const foo = function(bar) {
          if (bar) {
            return 1;
          } else {
            return 2;
          }
        };
      });`,
    },
    {
      code: `it('foo', () => {
        function foo(bar) {
          return foo ? bar : null;
        };
      });`,
    },
    {
      code: `it('foo', () => {
        function foo(bar) {
          if (bar) {
            return 1;
          } else {
            return 2;
          }
        };
      });`,
    },
```

cartogram:
@BPScott makes sense! may tackle normal functions in a follow-up though.

cartogram:
@BPScott Okay, regular functions are in https://github.com/Shopify/eslint-plugin-shopify/pull/331/commits/3d31d915407d663fb6ba11f1def1fc1b2eac84d2

BPScott (approved):
LGTM :+1:

PR #391, opened 2019-08-28
Upgrade Typescript

cartogram:
This PR upgrades Typescript and adds `@typescript-eslint/parser`.

### DIFF
--- a/lib/config/rules/prettier-typescript.js
+++ b/lib/config/rules/prettier-typescript.js
@@ -1,4 +1,5 @@
 module.exports = {
   // don't use lint rules, defer to prettier rules instead
-  'typescript/member-delimiter-style': 'off',
+  '@typescript-eslint/member-delimiter-style': 'off',
+  '@typescript-eslint/type-annotation-spacing': 'off',
 };
--- a/lib/config/rules/typescript.js
+++ b/lib/config/rules/typescript.js
@@ -1,57 +1,40 @@
-// See https://github.com/nzakas/eslint-plugin-typescript
+// See https://github.com/typescript-eslint/typescript-eslint
 
 module.exports = {
   // Enforce one space after the colon and zero spaces before the colon of a type annotation.
-  'typescript/type-annotation-spacing': ['error'],
-
+  '@typescript-eslint/type-annotation-spacing': ['error'],
   // Require explicit return types on functions and class methods
-  'typescript/explicit-function-return-type': 'off',
-
+  '@typescript-eslint/explicit-function-return-type': 'off',
   // Enforce accessibility modifiers on class properties and methods. (member-access from TSLint)
-  'typescript/explicit-member-accessibility': 'off',
-
+  '@typescript-eslint/explicit-member-accessibility': 'off',
   // Enforce interface names are prefixed. (interface-name from TSLint)
-  'typescript/interface-name-prefix': ['error', 'never'],
-
+  '@typescript-eslint/interface-name-prefix': 'off',
   // Enforce naming conventions for class members by visibility.
-  'typescript/member-naming': 'off',
-
+  '@typescript-eslint/member-naming': 'off',
   // Enforce /// <reference /> is not used. (no-reference from TSLint)
-  'typescript/no-triple-slash-reference': 'error',
-
+  '@typescript-eslint/no-triple-slash-reference': 'error',
   // Disallow generic Array constructors
-  'typescript/no-array-constructor': 'error',
-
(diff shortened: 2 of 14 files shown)


lib/config/rules/typescript.js
```diff
@@ -71,66 +54,73 @@ module.exports = {
...
+  '@typescript-eslint/ban-types': [
+    'error',
+    {
+      types: {
+        String: {message: 'Use string instead', fixWith: 'string'},
+        Boolean: {message: 'Use boolean instead', fixWith: 'boolean'},
+        Number: {message: 'Use number instead', fixWith: 'number'},
+        Object: {message: 'Use object instead', fixWith: 'object'},
+        Array: {message: 'Provide a more specific type'},
+      },
```
vsumner, line 93: ❤️

lib/config/rules/typescript.js
```diff
@@ -1,57 +1,40 @@
...
-
+  '@typescript-eslint/explicit-function-return-type': 'off',
   // Enforce accessibility modifiers on class properties and methods. (member-access from TSLint)
-  'typescript/explicit-member-accessibility': 'off',
-
+  '@typescript-eslint/explicit-member-accessibility': 'off',
   // Enforce interface names are prefixed. (interface-name from TSLint)
-  'typescript/interface-name-prefix': ['error', 'never'],
-
+  '@typescript-eslint/interface-name-prefix': 'off',
```
vsumner, line 11: It was never before and now it's off?

cartogram: Should be never 👍

cartogram, line 11:
```suggestion
  '@typescript-eslint/interface-name-prefix': ['error', 'never'],
```

package.json
```diff
@@ -71,13 +71,15 @@
     "prettier": "^1.17.0",
     "react": "^16.8.6",
     "react-dom": "^16.8.6",
-    "typescript": "~3.1.1"
+    "typescript": "^3.5.3"
```
vsumner, line 74: 🎉

lib/config/typescript.js
```diff
@@ -3,11 +3,17 @@ const merge = require('merge');
 module.exports = {
   extends: 'plugin:shopify/esnext',
 
-  plugins: ['typescript'],
+  plugins: [
+    '@typescript-eslint',
+    'eslint-comments',
+    'sort-class-members',
+    'import',
+    'shopify',
```
vsumner: Are these new or were they provided by the last version?

vsumner (approved):
Nice! This was causing issues with bumping webpack in `sewing-kit`.

lib/config/typescript.js
cartogram: They are not new. This array lists plugins whose rules are configured in the typescript rules config.

lib/config/rules/prettier-typescript.js
```diff
@@ -1,4 +1,5 @@
 module.exports = {
   // don't use lint rules, defer to prettier rules instead
-  'typescript/member-delimiter-style': 'off',
+  '@typescript-eslint/member-delimiter-style': 'off',
```
BPScott, line 3: I wonder if for this sort of thing we could leverage https://github.com/prettier/eslint-config-prettier/blob/master/%40typescript-eslint.js

Doesn't need to be done now but might be nice to be aware of.

lib/config/typescript.js
BPScott: We extend from shopify/esnext, which pulls in `sort-class-members` and`import` (see https://github.com/Shopify/eslint-plugin-shopify/blob/master/lib/config/esnext.js#L16)

shopify/esnext extends from core, which pulls in `shopify` and `eslint-comments` (see https://github.com/Shopify/eslint-plugin-shopify/blob/master/lib/config/core.js#L4)

What do we gain by being explicit here instead of relying on the transitive dependencies?

BPScott (approved):
Not tried this out but code looks sensible. Am excited to get rid of those override hacks!

Two small points inline that I'll defer to you on.

lib/config/typescript.js
cartogram: Probably nothing? I'll see what happens without the explicit plugins and go from there.

lib/config/rules/prettier-typescript.js
cartogram: https://github.com/Shopify/web-foundation/issues/104 i'll come back to it

PR #409, opened 2019-09-16
Enforce new lines between import groups

keyfer:
This enforces a new line between import groups, from this rule: https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md#newlines-between-ignorealwaysalways-and-inside-groupsnever

The reason I chose this approach is because it enforces what looks to be the side our codebase leans towards:

The effect of running with `always` (caps at 999)

The effect of running with `never`

With the addition of this rule the following are true:

**Invalid**

```js
import fs from 'fs';
import path from 'path';
import index from './';
import sibling from './foo';
```

**Valid**
```js
import fs from 'fs';
import path from 'path';

import index from './';

import sibling from './foo';
```

I attempted to see what this rule would cause if I also ran `--fix` on web, and this was the result: 
https://github.com/Shopify/web/pull/18382

### DIFF
--- a/lib/config/rules/import.js
+++ b/lib/config/rules/import.js
@@ -95,6 +95,7 @@ module.exports = {
     'error',
     {
       groups: ['builtin', 'external', 'internal', 'parent', 'sibling'],
+      'newlines-between': 'always',
     },
   ],
   // Enforce a newline after import statements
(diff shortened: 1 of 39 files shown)


BPScott (approved):

BPScott:
LGTM. As discussed in Slack I think that `always` is nearer to what usually we try and do.

Work out what's going on with the tests and add a CHANGELOG entry before merging

cartogram (approved):

cartogram:
@keyfer seems like something is up those tests that is not related to this PR. I am going to take a look.

cartogram:
@keyfer give this a rebase and I think you will be good to go.

keyfer:
Thanks @cartogram

PR #433, opened 2019-10-09
Update `@typescipt-eslint/*` packages

cartogram:
This PR brings us to the latest `@typescript-eslint/parser` and `@typescript-eslint/eslint-plugin`.

I've also added a `typescript-fast` config that can be used to augment the `typescript` config, but without the rules that require type checking.

More comments inline.

🎩 
To view the changes from this update applied to `quilt`, look at this branch https://github.com/Shopify/quilt/pull/1110/files. You can test yourself by  copying the files from this branch into your local `quilt/node-modules/eslint-plugin-shopify` and running `yarn lint`.

~~TODO: This will require changelog/readme updates. Will put those in once I get a general approval.~~

### Companion PRs

- https://github.com/Shopify/quilt/pull/1110
- https://github.com/Shopify/graphql-tools-web/pull/91
- https://github.com/Shopify/web-proving-ground/pull/36
- https://github.com/Shopify/web-rails-proving-ground/pull/254


lib/config/jest.js
```diff
@@ -14,7 +14,14 @@ module.exports = {
 
   rules: merge(require('./rules/jest'), {
     // Disallow vague words in test statements.
-    'shopify/jest/no-vague-titles': 'error',
+    // allowing "all"
+    'shopify/jest/no-vague-titles': [
+      'error',
+      {
+        allow: ['all'],
```
cartogram, line 21: This is basically disabled in our all our projects. Going to disable here, and eventually, we may likely just depreciate it.

lib/config/react.js
```diff
@@ -1,8 +1,6 @@
 const merge = require('merge');
 
 module.exports = {
-  extends: 'plugin:shopify/esnext',
```
cartogram, line 4: I don't think `react` should be a base config. I had a lot of trouble having a non-react typescript project without configs being unintentionally overwritten. This is a breaking change.

lib/config/rules/typescript.js
```diff
@@ -79,7 +75,13 @@ module.exports = {
   // Disallow the declaration of empty interfaces. (no-empty-interface from TSLint)
   '@typescript-eslint/no-empty-interface': 'off',
   // Requires using either T[] or Array<T> for arrays (array-type)
-  '@typescript-eslint/array-type': ['error', 'array'],
+  '@typescript-eslint/array-type': [
+    'error',
+    {
+      default: 'array-simple',
+      readonly: 'generic',
```
cartogram, line 94: I don't think we can use anything other than `generic` for the `readonly` property until we are on typescript 3.4 or greater.

lib/config/typescript.js
```diff
@@ -63,5 +79,16 @@ module.exports = {
...
     },
+    {
+      files: [
+        '**/test/**/*.ts',
+        '**/test/**/*.tsx',
+        '**/tests/**/*.ts',
+        '**/tests/**/*.tsx',
+      ],
+      rules: {
+        '@typescript-eslint/no-empty-function': 'off',
```
cartogram: We have a lot of `noop` functions in tests.

tests/lib/config/typescript.test.js
```diff
@@ -11,37 +11,5 @@ describe('config', () => {
         ),
       ).toBe('');
     }, 8000);
-
-    it('identifies import path issues', () => {
```
cartogram: typescript doesn't report path issues with the `esnext` change in the config.

lib/config/rules/typescript.js
cartogram: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html

lib/config/react.js
BPScott: So the new world is typescript is a base config, react is an augment? I like it!

Seems the [Provided configs](https://github.com/Shopify/eslint-plugin-shopify#provided-configurations) part of the readme needs some love - to add mention of the typescript core config, moving react to an augment, and removing typescript-prettier as that no longer exists

index.js
```diff
@@ -39,6 +39,7 @@ module.exports = {
     esnext: require('./lib/config/esnext'),
     react: require('./lib/config/react'),
     typescript: require('./lib/config/typescript'),
+    'typescript-fast': require('./lib/config/typescript-fast'),
```
BPScott: In what situations would we want new projects to use this over the typescript config? If we always want people to use the full-fat typescript config then is there much value exposing this?

lib/config/react.js
BPScott: Can we add some migration notes to make it clear that now people will need to specify a base config (esnext or typescript) if they're not already?

lib/config/typescript.js
BPScott: In Polaris we've got a several deliberate noop functions in our code too - many of which are default event handlers.

Can we keep this off for everything, not just tests?

lib/config/typescript.js
```diff
@@ -1,8 +1,6 @@
 const merge = require('merge');
 
 module.exports = {
-  extends: 'plugin:shopify/esnext',
```
BPScott, line 4: I think it makes sense to keep typescript as a base config - it's a foundational tech rather than a library you use.

BPScott (commented):
Comments inline.

I agree with making react be a augmentation, but I think typescript deserves its place as a base config

index.js
cartogram: The full-fat version will likely have a significant performance impact on web.

BPScott: Cool :) In that case if we keep it, what do you think to making it be a base-level config - so you import typescript or typescript-fast instead of needing both?

lib/config/typescript.js
cartogram: Yeah I am okay with this 👍

lib/config/rules/prettier-typescript.js
```diff
@@ -1,5 +1,11 @@
 module.exports = {
   // don't use lint rules, defer to prettier rules instead
+  '@typescript-eslint/quotes': 0,
```
lemonmade: `off`?

index.js
lemonmade: `typescript` needs to be the default one everyone will use. If there are additional type-based rules, put them in their own config that's additive with a name that makes it clear what they actually are, not just that they are slower.

lib/config/rules/typescript.js
```diff
@@ -123,4 +119,66 @@ module.exports = {
...
   // When adding two variables, operands must both be of type number or of type string
   '@typescript-eslint/restrict-plus-operands': 'error',
+  // Bans “// @ts-ignore” comments from being used
+  '@typescript-eslint/ban-ts-ignore': 'error',
+  // Enforce consistent brace style for blocks
+  '@typescript-eslint/brace-style': 'error',
+  // Enforces consistent usage of type assertions.
+  '@typescript-eslint/consistent-type-assertions': 'error',
+  // Consistent with type definition either interface or type
+  '@typescript-eslint/consistent-type-definitions': 'error',
```
lemonmade: Can we provide the explicit config where possible so the reader doesn't need to guess at the default?

lemonmade: Or at least a comment that clarifies what the choice is.

lib/config/rules/typescript.js
```diff
@@ -123,4 +119,66 @@ module.exports = {
...
+  // Requires that private members are marked as readonly if they're never modified outside of the constructor
+  '@typescript-eslint/prefer-readonly': 'error',
+  // Enforce the consistent use of either backticks, double, or single quotes
+  '@typescript-eslint/quotes': [
+    'error',
+    'single',
+    {avoidEscape: true, allowTemplateLiterals: true},
+  ],
+  // Enforce giving compare argument to Array#sort
+  '@typescript-eslint/require-array-sort-compare': 'error',
```
lemonmade: I don't think I've ever dine this...

lib/config/rules/typescript.js
```diff
@@ -123,4 +119,66 @@ module.exports = {
...
+  // Warns for any two overloads that could be unified into one by using a union or an optional/rest parameter
+  '@typescript-eslint/unified-signatures': 'error',
+  // Disallows awaiting a value that is not a Thenable
+  '@typescript-eslint/await-thenable': 'error',
+  // Disallow iterating over an array with a for-in loop
+  '@typescript-eslint/no-for-in-array': 'error',
+  // Avoid using promises in places not designed to handle them
+  '@typescript-eslint/no-misused-promises': 'error',
+  // Warns if a type assertion does not change the type of an expression
+  '@typescript-eslint/no-unnecessary-type-assertion': 'error',
```
lemonmade: some people will hate this rule

lib/config/typescript-fast.js
```diff
@@ -0,0 +1,32 @@
...
+    {
+      parser: '@typescript-eslint/parser',
+      parserOptions: {
+        ecmaVersion: 2017,
+        sourceType: 'module',
+      },
+      files: ['*.ts', '*.tsx'],
+      rules: {
+        // Disallows awaiting a value that is not a Thenable
+        '@typescript-eslint/await-thenable': 'off',
```
lemonmade: Is it possible to avoid using overrides? I feel like they cause a lot of problems in consumers because you need another override to override these rules. If only the parser bits are override, and the rules are not nested, does that break?

lib/config/typescript.js
```diff
@@ -35,15 +33,33 @@ module.exports = {
         'no-useless-constructor': 'off',
         'no-shadow': 'off',
         'no-use-before-define': 'off',
+        semi: 'off',
```
lemonmade, line 28: Same comment here, I feel overrides are making it difficult for people to later customize the config

package.json
```diff
@@ -78,8 +78,8 @@
     "prettier": "^1.14.0"
   },
   "dependencies": {
-    "@typescript-eslint/eslint-plugin": "^1.6.0",
-    "@typescript-eslint/parser": "^1.6.0",
+    "@typescript-eslint/eslint-plugin": "^2.3.3",
+    "@typescript-eslint/parser": "^2.3.3",
```
lemonmade, line 83: Can we make a note of specifically what TS versions our lint config effectively supports?

lib/config/typescript.js
```diff
@@ -1,8 +1,6 @@
 const merge = require('merge');
 
 module.exports = {
-  extends: 'plugin:shopify/esnext',
-
   plugins: ['@typescript-eslint'],
 
   settings: {
```
lemonmade: I think we are better off ditching the TS resolver too, anything that can be done in there can also be done with a bit more manual (but, generally, more correct) configuration of the default node resolver.

lemonmade: (comment for `import/resolver` setting below)

lemonmade (changes requested):

lib/config/rules/typescript.js
cartogram: Yeah same, I am going to turn it off.

lib/config/rules/typescript.js
cartogram: Why do you say that?

lemonmade: I think there's an argument to be made that `const foo: Foo = createFoo()` is valuable even if `createFoo` always returns `Foo` and the explicit type annotation is therefore unnecessary. Maybe less so for `const foo: string = 'foo';` but in general there are enough times that the explicitness is valuable that I think it's worth leaving it to dev discretion.

lib/config/rules/typescript.js
cartogram: That said, I think there was only one violation in quilt. We don't `sort` too often.

lib/config/typescript-fast.js
BPScott: I seem to recall we had something kinda like that before and it meant that typescript rules got applied to JS files, which caused problems because they didn't apply - e.g. a lint saying you should use `import` but you're writing a script to be ran in node where you only have access to `require()`.

The problems with needing overrides to trump overrides you mention have gone away in eslint 6 - they made a breaking change so that top-level rules now trump an override in an extended config - https://eslint.org/docs/user-guide/migrating-to-6.0.0#overrides-precedence. A parent config now always has precedence over everything in an extended config.

lib/config/typescript.js
BPScott: I believe that's going to mean people will have to redefine everything they specify in their tsconfig's `paths` config, per https://github.com/Shopify/eslint-plugin-shopify/pull/197

lib/config/typescript-fast.js
cartogram: I upgraded eslint in my tophat PR for this reason. https://github.com/Shopify/quilt/pull/1110/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R53

lib/config/typescript.js
cartogram: I think we leave the resolver change off for now. There is already a lot going on with this release.

lib/config/typescript-fast.js
lemonmade: Thank you both for the explanation ❤️

lib/config/typescript.js
lemonmade: Please consider it. I know they will need to redo the paths logic, but a) the paths logic breaks quilt, which necessitated the gross workaround in that PR, and related b) the way the TS resolver works is that anything with `paths` is considered an internal module (since it resolves all things to source), where we typically want it considered differently (internal is fine for the "include `./app` in the list of folders to resolve from" situation, but treating them as external is preferable for the "include `./packages` in the list of folders to resolve from " situation). Also, we will probably slowly move away from paths, at least for cases where TS project references are more appropriate.

cartogram: @lemonmade sounds good, will remove the TS resolver in this PR 👍

README.md
```diff
@@ -46,29 +46,33 @@ Shopify’s ESLint configs come bundled in this package. In order to use them, y
...
+  "extends": [
+    "plugin:shopify/esnext",
+    "plugin:shopify/jest"
+  ]
 }
 ```
 
-You can also add some "augmenting" configs on top of the "core" config by extending an array of linting configs. For example, this package provides a jest config, which can be added to the ESNext config with the following configuration file:
+
+Likewise, if you are using TypeScript and React, the following configuration extends the TypeScript base config with the React-specific rules provided by the React configuration file:
```
cartogram:
```suggestion
Likewise, if you are using TypeScript and React, the following example extends the TypeScript base configuration with the React-specific rules provided by the React configuration:
```

README.md
```diff
@@ -78,11 +82,25 @@ You can also add some "augmenting" configs on top of the "core" config by extend
 This plugin provides the following core configurations:
 
 - [esnext](lib/config/esnext.js): Use this for anything written with ES2015+ features.
-- [react](lib/config/react.js): Use this for React projects.
+- [typescript](lib/config/typescript.js): Use this for Typescript projects. The rules enabled in this confige do not require type-checking to run. To enable all Typescript rules, you must augment this config with the `typescript-requiring-type-checking` config mentioned below.
```
cartogram:
```suggestion
- [typescript](lib/config/typescript.js): Use this for Typescript projects. This configuration enables a subset of Typescript-specific that do not require type-checking to run. To enable all Typescript rules, you must augment this config with the `typescript-requiring-type-checking` config mentioned below.
```

README.md
```diff
@@ -78,11 +82,25 @@ You can also add some "augmenting" configs on top of the "core" config by extend
...
 - [es5](lib/config/es5.js): Use this for legacy projects.
 
 This plugin also provides the following tool-specific configurations, which can be used on top of the core configurations:
 
+- [typescript-requiring-type-checking](lib/config/typescript-requiring-type-checking.js) Use this config to augment the `typescript` config to enable all TypeScript rules, including those that require type checking. These rules are slower to run and and you will need to specify a path to your tsconfig.json file in the "project" property of "parserOptions". The following example would provide all of the TypeScript rules, assuming the tsconfig.json is in the same directory as you ESlint configuration.
```
cartogram:
```suggestion
- [typescript-requiring-type-checking](lib/config/typescript-requiring-type-checking.js) Use this to augment the `typescript` config and enable all TypeScript rules, including those that require type checking. These rules are slower to run, and you will need to specify a path to your tsconfig.json file in the "project" property of "parserOptions". The following example would provide all of the TypeScript rules, assuming the tsconfig.json is in the same directory as your ESlint configuration.
```

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-### Breaking Changes
+### Using `@typescript-eslint/parser` and `@typescript-eslint/eslint-plugin`
```
lemonmade, line 5: This reads weirder to me than "Breaking changes" /shrug

CHANGELOG.md
```diff
@@ -22,6 +22,12 @@ Will become:
 }
 ```
 
+### Config Changes
+
+- **Breaking Change** The `plugin:shopify/react` is no longer a core config and must augment one of the `plugin:shopify/typescript` or `plugin:shopify/esnext` configs
```
lemonmade: Will people know what this even means for them? Can we show them an example config diff?

README.md
```diff
@@ -46,29 +46,33 @@ Shopify’s ESLint configs come bundled in this package. In order to use them, y
...
 
-If working on an ES5 project, extend the ES5 version of the configuration:
+You can also add some "augmenting" configs on top of the "core" config by extending an array of linting configs. For example, the following configuration would provide a base ESNext config that is augmented by a jest config:
 
 ```json
 {
-  "extends": "plugin:shopify/es5"
+  "extends": [
+    "plugin:shopify/esnext",
+    "plugin:shopify/jest"
```
lemonmade: I'd rather we use another config that should apply universally (IMO most projects should have the jest rule isolated to an `override`)

README.md
```diff
@@ -78,11 +82,25 @@ You can also add some "augmenting" configs on top of the "core" config by extend
...
 
 - [esnext](lib/config/esnext.js): Use this for anything written with ES2015+ features.
-- [react](lib/config/react.js): Use this for React projects.
+- [typescript](lib/config/typescript.js): Use this for Typescript projects. This configuration enables a subset of Typescript-specific that do not require type-checking to run. To enable all Typescript rules, you must augment this config with the `typescript-requiring-type-checking` config mentioned below.
 - [es5](lib/config/es5.js): Use this for legacy projects.
 
 This plugin also provides the following tool-specific configurations, which can be used on top of the core configurations:
 
+- [typescript-requiring-type-checking](lib/config/typescript-requiring-type-checking.js) Use this to augment the `typescript` config and enable all TypeScript rules, including those that require type checking. These rules are slower to run, and you will need to specify a path to your tsconfig.json file in the "project" property of "parserOptions". The following example would provide all of the TypeScript rules, assuming the tsconfig.json is in the same directory as your ESlint configuration.
```
lemonmade: I would drop `-requiring` from the name

lib/config/rules/typescript.js
```diff
@@ -11,18 +11,23 @@ module.exports = {
   '@typescript-eslint/interface-name-prefix': 'off',
```
lemonmade, line 11: I noticed in one of the projects that I was getting a lint error for both the TypeScript `camelcase` rule and the Babel one. Is that resolved by this PR?

package.json
lemonmade: Was this done?

lemonmade (approved):
Seems good to me now 👍

README.md
BPScott: Alternative naming bikeshedding: `typescript-extended` or `typescript-advanced`?

tests/fixtures/typescript-imports/.eslintrc.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
   plugins: ['self'],
-  extends: ['plugin:self/typescript', 'plugin:self/node'],
+  extends: ['plugin:self/node', 'plugin:self/typescript'],
```
BPScott: This doesn't seem right - typescript is a core config and node is an augment so the ordering should be extending from typescript then node

BPScott (commented):
Not ran but looking good!

Agree with Chris's request for code examples in the changelog

Some petty bikeshedding about the name of the on-steroids TS config but not a big deal.

The README still mentions the `typescript-prettier` augment config which no longer exists - can that be deleted?

lib/config/rules/typescript.js
cartogram: I remember noticing that at one point as well. It is solved by this PR as shown in the quilt companion PR here: https://github.com/Shopify/quilt/pull/1110/files#diff-d8cfb868aeea19ce239de10dcfdbd4c2R55

tests/fixtures/typescript-imports/.eslintrc.js
cartogram: oof, good catch!

PR #471, opened 2019-11-14
Use Overrides in GraphQL Config

cartogram:
Look at the `Changelog.md` for a description of the changes.

One downside to this change is that I think it makes the ability to check `gql` template literals impossible. IMO this is not a problem because we don't consider those a best practise, but maybe this is too autocratic? 🤷‍♂


lemonmade (approved):
Seems like a good approach to me 👍

PR #498, opened 2019-12-04
Support arrow functions in react-no-multiple-render-methods

vhpoet:
Title is self explanatory.

### DIFF
--- a/lib/rules/react-no-multiple-render-methods.js
+++ b/lib/rules/react-no-multiple-render-methods.js
@@ -24,7 +24,7 @@ module.exports = {
     let isES6Component = false;
 
     function report(node) {
-      const name = (node.key && node.key.name) || (node.parent && node.parent.key && node.parent.key.name);
+      const name = getMethodName(node);
 
       context.report({
         node,
@@ -45,7 +45,7 @@ module.exports = {
 
         report(node);
       },
-      
+
       ArrowFunctionExpression(node) {
         if (!isES6Component || isRenderMethod(node)) {
           return;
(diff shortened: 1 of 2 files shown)


GoodForOneFare:
Thanks for the contribution, Vahe! ❤️ @cartogram could you take a look when you have a moment, please?

cartogram:
@vhpoet do mind adding a test for this?

lib/rules/react-no-multiple-render-methods.js
```diff
@@ -47,10 +45,19 @@ module.exports = {
...
+
+        report(node);
+      },
     };
   }),
 };
 
-function isRenderMethod({key: {name}}) {
+function isRenderMethod(node) {
+  const name = (node.key && node.key.name) || (node.parent && node.parent.key && node.parent.key.name);
```
cartogram: let's break this out into a small utility function. Something like:
```
getMethodName(node) {
  if (node.key) {
     return node.key.name
  }

  return node.parent && node.parent.key && node.parent.key.name;
}
```

lib/rules/react-no-multiple-render-methods.js
```diff
@@ -47,10 +45,19 @@ module.exports = {
 
         report(node);
       },
+      
+      ArrowFunctionExpression(node) {
```
cartogram, line 49: You are going to need to add a few tests for this. At the least I'd like to see it catch something like:
```
      code: `class Button extends React.Component {
        renderArrowFunction = () => {}
        renderBar = () => {}
        render() {}`
```

I think the error type will be different for these, so you will also need to modify the `error` utility to accept a `type` argument. Something like:

```
function error(memberName, type = 'MethodDefinition') {
  return {
    type,
    message: `Don’t use multiple render methods in a single component; they generally make your component harder to read. Instead break ${memberName} out into its own component and render it inside this one.`,
  };
}
```
then you should be able to do the following for the error of the above snippet:
```
errors: [
  error('renderArrowFunction', 'ArrowFunctionExpression'),
  error('renderBar', 'ArrowFunctionExpression'),
],
```

Happy to clarify anything you need.

cartogram (changes requested):
Thanks for this contribution! Will take a look again once these changes are in 🙂

alexandcote:
I did the rework you asked @cartogram. Let me know if everything is 👍🏼.

Thanks again @vhpoet 🤟🏼

BPScott (approved):
Not tried this but the test LGTM

PR #514, opened 2020-01-12
Use word boundaries for vague terms separation

YannCedric:
### Problem
The `no-vague-titles` rule on test statements is perhaps too general. 
For most keywords, the rule matches sequences instead of words.
### Example
If I use a variable that contains a blacklisted word like `someVariableWithEveryInIt` in my title , I get a `no-vague-titles` error because of **every**, even though I technically did not use the word **every** in my title.
### Fix
Change the regular expressions to match words instead of sequences.
Instead of checking if the test statement contains a **sequence**, use [regex's word boundaries](https://www.regular-expressions.info/wordboundaries.html) to check if the string contains a **word**.
>Similar to the way **all** is configured, only difference is the optional prefix/suffix.

### DIFF
--- a/lib/rules/jest/no-vague-titles.js
+++ b/lib/rules/jest/no-vague-titles.js
@@ -7,14 +7,15 @@ const {
 } = require('../../utilities/jest');
 
 const VAGUE_TERMS = {
-  should: /should/i,
-  correct: /correct/i,
-  appropriate: /appropriate/i,
+  correct: /\b(in)?correct(ly)?\b/i,
+  appropriate: /\b(in)?appropriate(ly)?\b/i,
+  properly: /\b(in)?proper(ly)?\b/i,
+  necessary: /\b(un)?necessary(ly)?\b/i,
+
+  descriptive: /\bdescriptive\b/i,
+  should: /\bshould\b/i,
   all: /\ball\b/i,
-  properly: /properly/i,
-  every: /every/i,
-  descriptive: /descriptive/i,
-  necessary: /necessary/i,
+  every: /\bevery\b/i,
 };
 
 module.exports = {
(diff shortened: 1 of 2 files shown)


cartogram (approved):
💯 Thanks Yann!

lib/rules/jest/no-vague-titles.js
```diff
@@ -7,14 +7,15 @@ const {
...
+  necessary: /\b(un)?necessary(ly)?\b/i,
+
+  descriptive: /\bdescriptive\b/i,
+  should: /\bshould\b/i,
   all: /\ball\b/i,
-  properly: /properly/i,
-  every: /every/i,
-  descriptive: /descriptive/i,
-  necessary: /necessary/i,
+  every: /\bevery\b/i,
```
clauderic, line 18: nit: this was catching vague words like `everything` and `everywhere` before. I'm not sure about `everywhere`, but I think having `everything` in a test name should probably be classified as vague.

lib/rules/jest/no-vague-titles.js
```diff
@@ -7,14 +7,15 @@ const {
 } = require('../../utilities/jest');
 
 const VAGUE_TERMS = {
-  should: /should/i,
-  correct: /correct/i,
-  appropriate: /appropriate/i,
+  correct: /\b(in)?correct(ly)?\b/i,
+  appropriate: /\b(in)?appropriate(ly)?\b/i,
+  properly: /\b(in)?proper(ly)?\b/i,
```
clauderic, line 12: typo: 

```suggestion
  properly: /\b(im)?proper(ly)?\b/i,
```

lib/rules/jest/no-vague-titles.js
YannCedric: Yeah, I agree. I think everywhere makes sense to catch as well.
In that case, I think simply enumerating `everything` and `everywhere` is better (?)

clauderic: I'm not sure whether `everywhere` makes sense considering we extensively use that for legitimate naming at Shopify (Sections Everywhere for example), but yeah I'd add `everything` as a separate case that is enumerated

PR #515, opened 2020-01-13
Fix typos and whitelist keyword `everything`

YannCedric:
Following up on https://github.com/Shopify/eslint-plugin-shopify/pull/514, fixing a typo and whitelisting the keyword `everything`, which we deem to also be vague in a test title. This keyword was previously being matched by the `every` regex, so this shouldn't raise any new errors

### DIFF
--- a/lib/rules/jest/no-vague-titles.js
+++ b/lib/rules/jest/no-vague-titles.js
@@ -9,13 +9,14 @@ const {
 const VAGUE_TERMS = {
   correct: /\b(in)?correct(ly)?\b/i,
   appropriate: /\b(in)?appropriate(ly)?\b/i,
-  properly: /\b(in)?proper(ly)?\b/i,
+  properly: /\b(im)?proper(ly)?\b/i,
   necessary: /\b(un)?necessary(ly)?\b/i,
 
   descriptive: /\bdescriptive\b/i,
   should: /\bshould\b/i,
   all: /\ball\b/i,
   every: /\bevery\b/i,
+  everything: /\beverything\b/i,
 };
 
 module.exports = {
--- a/lib/rules/jest/no-vague-titles.js
+++ b/lib/rules/jest/no-vague-titles.js
@@ -15,8 +15,7 @@ const VAGUE_TERMS = {
   descriptive: /\bdescriptive\b/i,
   should: /\bshould\b/i,
   all: /\ball\b/i,
-  every: /\bevery\b/i,
-  everything: /\beverything\b/i,
+  every: /\bevery(thing)?\b/i,
 };
 
 module.exports = {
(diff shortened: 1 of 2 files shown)


tests/lib/rules/jest/no-vague-titles.test.js
```diff
@@ -91,7 +91,7 @@ ruleTester.run('no-vague-titles', rule, {
       code: `someFunction('Necessary')`,
     },
     {
-      code: `someFunction('necessaryly')`,
+      code: `someFunction('necessarily')`,
```
clauderic, line 94: Good catch, but then the regex above is wrong, no?  `necessary: /\b(un)?necessary(ly)?\b/i,`

YannCedric: Indeed, it would be wrong.
Once again, I'll just enumerate the two rules 👌

clauderic: I'm a bit confused that the tests were passing though 🤔

lib/rules/jest/no-vague-titles.js
```diff
@@ -9,13 +9,15 @@ const {
 const VAGUE_TERMS = {
   correct: /\b(in)?correct(ly)?\b/i,
   appropriate: /\b(in)?appropriate(ly)?\b/i,
-  properly: /\b(in)?proper(ly)?\b/i,
-  necessary: /\b(un)?necessary(ly)?\b/i,
+  properly: /\b(im)?proper(ly)?\b/i,
+  necessary: /\b(un)?necessary\b/i,
+  necessarily: /\b(un)?necessarily\b/i,
```
clauderic: This could actually be combined: `/\b(un)?necessar(y|ily)\b/i`

tests/lib/rules/jest/no-vague-titles.test.js
YannCedric: I believe they that section is related to calls to "non test" functions like `it()`, `describe()` and so on. (Hence the `someFunction()`?)
I added tests under `invalid` cases to validate negative scenarios instead.

lib/rules/jest/no-vague-titles.js
YannCedric: Indeed

lib/rules/jest/no-vague-titles.js
```diff
@@ -9,13 +9,14 @@ const {
...
-  properly: /\b(in)?proper(ly)?\b/i,
-  necessary: /\b(un)?necessary(ly)?\b/i,
+  properly: /\b(im)?proper(ly)?\b/i,
+  necessary: /\b(un)?necessar(y|ily)\b/i,
 
   descriptive: /\bdescriptive\b/i,
   should: /\bshould\b/i,
   all: /\ball\b/i,
   every: /\bevery\b/i,
+  everything: /\beverything\b/i,
```
clauderic: nit: I suppose this could also be combined 🤷‍♂ 

`every: /\bevery(thing)?\b/i,`

clauderic (approved):
Looks great, thanks for cleaning this up 💯

lib/rules/jest/no-vague-titles.js
YannCedric: Totally 😅

PR #517, opened 2020-01-20
[typescript/prefer-pascal-case-enum] Fix an issue with undefined name

### LINKED ISSUE #516: Bug with [typescript/prefer-pascal-case-enum] and string keys
We tried to bump eslint-plugin-shopify to the last version (34.0.0) in our project and we faced an issue.

<details>
<summary>Stacktrace</summary>

````
TypeError: Cannot read property 'replace' of undefined
Occurred while linting /Users/alexandrecote/src/github.com/Shopify/online-store-web/app/components/RichTextEditor/types.ts:61
    at /Users/alexandrecote/src/github.com/Shopify/online-store-web/node_modules/pascal-case/node_modules/no-case/dist/index.js:36:58
    at Array.reduce (<anonymous>)
    at replace (/Users/alexandrecote/src/github.com/Shopify/online-store-web/node_modules/pascal-case/node_modules/no-case/dist/index.js:36:15)
    at Object.noCase (/Users/alexandrecote/src/github.com/Shopify/online-store-web/node_modules/pascal-case/node_modules/no-case/dist/index.js:14:26)
    at pascalCase (/Users/alexandrecote/src/github.com/Shopify/online-store-web/node_modules/pascal-case/dist/index.js:20:22)
    at isPascalCase (/Users/alexandrecote/src/github.com/Shopify/online-store-web/node_modules/eslint-plugin-shopify/lib/rules/typescript/prefer-pascal-case-enums.js:42:19)
    at TSEnumMember (/Users/alexandrecote/src/github.com/Shopify/online-store-web/node_modules/eslint-plugin-shopify/lib/rules/typescript/prefer-pascal-case-enums.js:28:14)
    at /Users/alexandrecote/src/github.com/Shopify/online-store-web/node_modules/eslint/lib/linter/safe-emitter.js:45:58
    at Array.forEach (<anonymous>)
    at Object.emit (/Users/alexandrecote/src/github.com/Shopify/online-store-web/node_modules/eslint/lib/linter/safe-emitter.js:45:38)
````

</details>

After some investigation, we have discovered that having a string as key in an enum cause the error. No `name` is provided [here](https://github.com/Shopify/eslint-plugin-shopify/blob/master/lib/rules/typescript/prefer-pascal-case-enums.js#L42).

You can reproduce this bug in the test environment with 
```
{
  code: `enum Example {'foo' = 'bar'}`,
  errors: [],
},
```

I suspect [this PR](https://github.com/Shopify/eslint-plugin-shopify/pull/500) introduced the bug. The old pascal-case look for `undefined` / `null` value.

You can check here. Updating the lib to the last version causes the same issue if you pass undefined.
[Edit pascal-case tests](https://codesandbox.io/s/silly-galileo-sb0cw?fontsize=14&hidenavigation=1&theme=dark)

Now the question is should ignore the check if the name is undefined or we should find why the name is set to `undefined` 🤔. This is the main reason why I didn't add the `if`.

#### Example of code in our codebase.

```ts
export enum ImageSize {
  Small = '100x100',
  '1024x1024' = '1024x1024'
}
```

alexandcote:
Fix: #516

1. We use the node `value` when the `name` is `undefined`. When the key of the enum is declared as a string, the `name` is `undefined`
2. We check if the identifier is defined before we pass the value to the `pascalCase` function.

You can find more context in the issue.

### DIFF
--- a/tests/lib/rules/typescript/prefer-pascal-case-enums.test.js
+++ b/tests/lib/rules/typescript/prefer-pascal-case-enums.test.js
@@ -8,7 +8,6 @@ const ruleTester = new RuleTester({
 
 function errorWithName(name) {
   return {
-    type: 'Identifier',
     message: `Enum '${name}' should use Pascal case.`,
   };
 }
@@ -43,6 +42,10 @@ ruleTester.run('prefer-pascal-case-enums', rule, {
     {
       code: `enum SortOrder {MOSTRECENT, least_recent, Newest, Oldest}`,
       errors: [errorWithName('MOSTRECENT'), errorWithName('least_recent')],
+    },
+    {
+      code: `enum Example {'foo' = 'bar', '1024x1024' = '1024x1024', Oldest}`,
+      errors: [errorWithName('foo')],
     },
   ],
 });
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 - remove `no-vague-titles` because the rule was adopted into `eslint-plugin-jest`'s `valid-title` rule. See [the `valid-title` documentation](https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/valid-title.md#disallowedwords)
+- Fixed an issue with `typescript/prefer-pascal-case-enum` when you had enum with key as string. ([517](https://github.com/Shopify/eslint-plugin-shopify/pull/517))
 
 ## [34.0.1] - 2019-01-13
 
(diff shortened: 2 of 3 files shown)


tests/lib/rules/typescript/prefer-pascal-case-enums.test.js
```diff
@@ -8,7 +8,6 @@ const ruleTester = new RuleTester({
 
 function errorWithName(name) {
   return {
-    type: 'Identifier',
```
alexandcote, line 11: I had to remove this check because when the key is a string, the type of the error is `Literal`

cartogram: That makes sense

cartogram (approved):
Sorry it took me so long to get to this :( 

LGTM! Could you please add a Changelog entry?  Let's work together to do a patch release once this is merged. I'd like to get this out to folks quickly.  Thanks again ❤️

PR #518, opened 2020-02-05
Add graphql env option

CameronGorrie:
GraphQL lint rules were not working without the env configuration. I was able to confirm this fix works for multiple schemas defined in a `graphqlconfig` file.

### DIFF
--- a/lib/config/rules/graphql.js
+++ b/lib/config/rules/graphql.js
@@ -1,8 +1,8 @@
 // uses graphql-config for schema mapping
 module.exports = {
   'graphql/capitalized-type-name': 'off',
-  'graphql/named-operations': 'error',
-  'graphql/no-deprecated-fields': 'error',
+  'graphql/named-operations': ['error', {env: 'literal'}],
+  'graphql/no-deprecated-fields': ['error', {env: 'literal'}],
   'graphql/template-strings': ['error', {env: 'literal'}],
   'graphql/required-fields': 'off',
 };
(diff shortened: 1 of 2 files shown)


lib/config/rules/graphql.js
```diff
@@ -1,8 +1,8 @@
 // uses graphql-config for schema mapping
 module.exports = {
   'graphql/capitalized-type-name': 'off',
-  'graphql/named-operations': 'error',
-  'graphql/no-deprecated-fields': 'error',
+  'graphql/named-operations': ['error', {env: 'literal'}],
```
donnielrt, line 4: Curious why we didn't use [env: 'apollo'](https://github.com/apollographql/eslint-plugin-graphql#common-options) here? Is that for when Apollo Engine is used or something?

CameronGorrie: Docs aren't super clear. The difference is that we use [`.graphql` files (literal)](https://github.com/apollographql/eslint-plugin-graphql#graphql-literal-files) rather than some flavour of [template string queries in JS](https://github.com/apollographql/eslint-plugin-graphql#identity-template-literal-tag). I had originally expected the rules to work out of the box since 'apollo' is the default and we use Apollo, but 'literal' is what actually works.

donnielrt: Right, that makes a lot of sense. I suppose `apollo` in JS uses the [graphql-tag](https://github.com/apollographql/graphql-tag) library, which is why they have a separate option for it. :neat:

donnielrt (approved):
:shipit:

